PR #59, opened 2015-01-08
Deprecation warning fixed

philipp-classen:


etr:
Thank you. That was very old code lying in there :).

PR #60, opened 2015-01-18
Remove useless code

a-pavlov:
vector guarantees the placement of element in memory like in array


etr:
Seems legit. Thank you!

PR #64, opened 2015-02-23
Return the status from the call webserver::register_resource

craigminihan:


src/httpserver/webserver.hpp
```diff
@@ -95,19 +95,20 @@ class webserver
...
         **/
         template <typename T>
-        void register_resource(const std::string& resource,
+        bool register_resource(const std::string& resource,
                 http_resource<T>* res, bool family = false
         )
         {
             details::http_resource_mirror hrm(res);
-            register_resource(resource, hrm, family);
+            return register_resource(resource, hrm, family);
```
craigminihan: I was trying to track down an issue with resource registration when I noticed the status wasn't being returned here.

craigminihan:
Moved this on to an issue branch

Issue #65, opened 2015-02-23
A registered root is never invoked


craigminihan:
Consider the following:

```c++
  ws.register_resource("/xyz", &xyz_r, false);
  ws.register_resource("/", &root_r, false);
```

Sending a request for /xyz works fine. Sending a request for / always fails with a 404. Setting `family` to `true` then sends unwanted requests like `/abc` to `root_r` which isn't right either.

The problem is due to path handling in `webserver::finalize_answer`, specifically the lines:
```c++
  const char* st_url = mr->standardized_url->c_str();
               
  fe = registered_resources_str.find(st_url);
  if(fe == registered_resources_str.end())
```

Sending a `curl http://localhost:8080/` yields `mr->standardized_url` == "" and not "/" as you might expect. `registered_resources_str` contains "/xyz" and "/" so the `find` always misses resulting in the 404. Ideally `st_url` should be set to "/" when the length of standardized_url is zero.

etr:
Solved by related pull request

PR #66, opened 2015-02-23
standardize_url doesn't remove the root / now

craigminihan:
This addresses issue #65. The solution here is to prevent http_utils::standardize_url() from removing the trailing / if the uri is only 1 character long.


etr:
Thanks for this change - it looks positive. The only thing I would add is a test that assets the new behaviour.

craigminihan:
Added test cases

etr:
Everything seems good. Thank you very much for this!

PR #68, opened 2015-02-24
Return the status from the call webserver::register_resource

craigminihan:


etr:
Everything good here. I am ok with the change of interface especially if it turns to be useful.

PR #72, opened 2015-04-02
http_response: Change cycle_callback_ptr so that buffer can be modified

vozhyk-:
With this, the callback can modify the buffer to put the response data there. 'max' should be provided to the callback as well, so that it doesn't write over the end of the buffer.
A char * can't just be passed to the callback as std::string without any conversion.

Also, http_response_builder::deferred_response and cycle_callback_ptr could use some documentation. Some can be adapted from [1] and [2].
[1] https://www.gnu.org/software/libmicrohttpd/manual/libmicrohttpd.html#microhttpd_002dresponse-create
[2] https://www.gnu.org/software/libmicrohttpd/manual/libmicrohttpd.html#microhttpd_002dcb

### DIFF
--- a/src/http_response.cpp
+++ b/src/http_response.cpp
@@ -207,7 +207,7 @@ namespace details
 
 ssize_t cb(void* cls, uint64_t pos, char* buf, size_t max)
 {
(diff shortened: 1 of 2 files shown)


etr:
Definitely a positive change.

Thanks!

PR #73, opened 2015-04-10
Make cycle_callback a functor (+ rename cycle_callback -> data_callback)

vozhyk-:
There can be a need to preserve some state across calls to the callback. With this change, it becomes possible to do this using a subclass of data_callback [[1]](https://github.com/vozhyk-/megahttp/blob/bd704573181e8f2b84ec8db280b951d709808817/src/megahttp_resource.cpp#L35).
I'm not sure if it's the best way to do it. The same can be done with a C++11 capturing lambda (and supported with templates on `libhttpserver` side).
Also, making it an alternative to a simple function pointer instead of replacing it looks like a better idea, but I don't know which way to do it is the best.


src/httpserver/http_response.hpp
```diff
@@ -62,7 +63,23 @@ class bad_caching_attempt: public std::exception
...
+/**
+ * Callback called to obtain data from a deferred_response
+**/
+class data_callback
+{
+public:
+    /**
+     * @param buf Buffer to copy data to
+     * @param max Maximum number of bytes allowed to be copied
+     * @return number of bytes copied, or
```
vozhyk-: This can be improved by mirroring MHD documentation

etr:
This is definitely interesting.
Give me some time to try possible different approaches and I'll come back to you.

etr:
Added to the repo: https://github.com/etr/libhttpserver/pull/145

PR #81, opened 2015-09-08
Add CMake module to Makefile.am to be installed

vozhyk-:
The CMake module wasn't installed by `make install` before.


etr:
Everything seems ok. Merged

PR #82, opened 2015-10-02
FIX: regex-matching was broken if (!family_url)

4picht:
When creating endpoints like
    ws.register_resource("/hello/{test|[0-9]+}", &hwr, false);
the match never succeeded, because the regex was matched against the modded request URL which contained '^' and '$'.


etr:
Thanks for spotting the bug and fixing it

PR #83, opened 2015-10-02
Do not call abort, throw an exception instead

4picht:
Calling abort is a really bad idea inside a library, because the whole application will die when just the HTTP server module has a problem.


etr:
Very positive. Thanks for the patch.

PR #84, opened 2015-10-05
webserver: stop leaking pthreads on startup failure

4picht:
Since the program no longer aborts when a failure occures, all threads started need to be stopped again.


etr:
I will add some test to assess this behavior - merging this. Thanks for the fix.

PR #91, opened 2016-03-14
FIX: logic error in refactoring

pnxs:
with commit cccacb6c91da57ac2934c4df916ff24f6e2ef793 the logic was broken at
two points:

For the case "(parts[i] != "") && (parts[i][0] != '{') and ! first"
url_modded was set to "/" + url_modded + parts[i] instead of url_modded + "/" + parts[i]

The second error was subtracting variable 'bar' when bar == string::npos


etr:
Thanks for spotting

PR #93, opened 2016-04-16
Fix warnings about const results of functions

wlandry:
This fixes some warnings I get from gcc 4.9.2 like

/home/boo/ipac/ttmp/libhttpserver-bin/include/httpserver/http_utils.hpp:306:24: warning: type qualifiers ignored on function return type [-Wignored-qualifiers]
     const int weight() const


etr:
Looks good - thanks for taking care of this

PR #94, opened 2016-04-18
Parse POST's with format of formdata as well as urlencoded.

wlandry:
This enables the underlying capability of libmicrohttpd to parse POST's with the format of "multipart/form-data".  This is useful for parsing simple curl requests like

curl -F "foo=bar" http://localhost/


etr:
Looks good. Thanks for the support!

PR #95, opened 2016-04-18
Handle POST's larger than 1024 bytes correctly.

wlandry:
webserver::post_iterator unconditionally sets a key to the last value parsed.  But if the post is larger than 1024 bytes, it will parse it incrementally.  That means that the key will be set to the last part, instead of the whole POST.  This fix prepends the existing value if any.


wlandry:
I have added a facility to limit the size of POST's, so that attackers can not upload terabyte files that crash the server.  The default is unlimited.

src/httpserver/http_request.hpp
```diff
@@ -442,7 +444,7 @@ class http_request
         **/
         void set_arg(const std::string& key, const std::string& value)
         {
-            this->args[key] = value;
+            this->args[key] = value.substr(0,content_size_limit);
```
etr: In this case, it feels like throwing an exception would be more transparent - at least the library consumer will be notified that it happened and will avoid using partial data as argument unless they want to.

wlandry: I did think of that. However, if I threw an exception, the functions `webserver::build_request_args`, and `webserver::post_iterator` would have to catch it immediately. They are invoked by libmicrohttpd, so they can not propagate exceptions.  Then we would still need a way to indicate that too many characters were POST'ed.

With the solution I gave, the user can check whether each of the args has size >= content_size_limit. We could even add a simple function to check overflow.  A C++98 version would be something like.

```c++
bool overflow() const
{
    for (std::map<std::string, std::string, http::arg_comparator>::const_iterator arg=args.begin();
          arg!=args.end(); ++arg)
    {
        if (arg->second.size()>=content_size_limit)
          return true;
    }
    return false;
}
```

Let me know what you think.

etr: I see.

I think, that having this kind of semantic, a utility such as the one you are describing could be very useful.

wlandry: It turns out to be enough to just check the size of `content`.  I added a function `content_too_large()`.

I also had to roll back pellucide's modification since it conflicted with checking the size.

etr:
Thanks. All looks good overall.

PR #96, opened 2016-04-18
Increase memory limit for POST processing to 32kb

wlandry:
This patch increases the buffer that POST processing uses to 32 kb, which is the same as  #MHD_POOL_SIZE_DEFAULT.  This speeds up processing large POST's.

### DIFF
--- a/src/webserver.cpp
+++ b/src/webserver.cpp
@@ -673,9 +673,10 @@ int webserver::bodyfull_requests_answer_first_step(
         )
     )
     {
+        const size_t post_memory_limit (32*1024);  // Same as #MHD_POOL_SIZE_DEFAULT
         mr->pp = MHD_create_post_processor (
                 connection,
-                1024,
+                post_memory_limit,
                 &post_iterator,
                 mr
         );


src/webserver.cpp
```diff
@@ -673,9 +673,10 @@ int webserver::bodyfull_requests_answer_first_step(
         )
     )
     {
+        const size_t post_memory_limit (32*1024);  // Same as #MHD_POOL_SIZE_DEFAULT
         mr->pp = MHD_create_post_processor (
                 connection,
-                1024,
+                post_memory_limit,
```
etr: Would it be possible to use the constant #MHD_POOL_SIZE_DEFAULT directly so that in case it changes in libmicrohttpd also libhttpserver will immediately benefit from the change?

wlandry: Unfortunately, MHD_POOL_SIZE_DEFAULT is only defined in src/microhttpd/daemon.c.  So it is not visible to library users.

etr: I see - not much we can do then.

etr:
Merging.

PR #99, opened 2016-07-15
Update debian/ install files to handle the latest multiarch configuration

wlandry:
With multiarch, debian puts library and pkgconfig files into `/usr/lib/x86_64-linux-gnu/` (for amd64), not `/usr/lib/`.


etr:
Merged.

Thanks for the help!

PR #105, opened 2016-11-07
fixed a problem that causes a memory leak. We are passing user, pass,

rdiazmartin:
We are passing user, pass and digested_user as char pointers by value which causes the original
pointers be always null when they are going to be free in
complete_request, now we pass them by reference and everything works
fine


etr:
Looks alright. Merging it.

PR #113, opened 2017-04-26
Reserve space for the c-string null termination

martamoreton:
Reserve space for the c-string null termination


etr:
Implemented in a recent commit. Thanks a lot for notifying.

PR #116, opened 2017-11-07
Manage ipv6 in get_ip_str function

GMercat:
sockaddr_in (IPv4) is the same size as sockaddr, which is why getnameinfo() is working for IPv4. But sockaddr_in6 (IPv6) is larger than sockaddr, which is why getnameinfo() fails.

### DIFF
--- a/src/http_utils.cpp
+++ b/src/http_utils.cpp
@@ -255,10 +255,6 @@ void get_ip_str(
         {
             result = to_ret;
         }
-        else
-        {
-            result.clear();
-        }
     }
 }
 
--- a/src/http_utils.cpp
+++ b/src/http_utils.cpp
@@ -244,9 +244,21 @@ void get_ip_str(
 {
     if(sa)
     {
+        int addrlen = sizeof(sockaddr_in);
+        if (AF_INET6 == sa->sa_family)
+        {
+            addrlen = sizeof(sockaddr_in6);
+        }
+
         char to_ret[NI_MAXHOST];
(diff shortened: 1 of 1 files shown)


src/http_utils.cpp
```diff
@@ -244,9 +244,21 @@ void get_ip_str(
...
+        }
+
         char to_ret[NI_MAXHOST];
-        getnameinfo(sa, sizeof (struct sockaddr), to_ret, NI_MAXHOST, NULL, 0, NI_NUMERICHOST);
-        result = to_ret;
+        if (0 == getnameinfo(sa, addrlen, to_ret, NI_MAXHOST, NULL, 0, NI_NUMERICHOST))
+        {
+            result = to_ret;
+        }
+        else
```
etr: The rest looks good but I am not sure why you are clearing the result here.

GMercat: I can remove the `else` statement, no prob.

Because it's how i code 😉 For this type of function, i prefer return a cleared object instead of do nothing about it. Users of this function have always  a coherent object (setted or cleared). It's just my point of view.

Tell me what do you want, remove or not?

etr: And you are probably right. For now, remove it though for consistency with the rest of the codebase.

I am anyhow disagreeing with my old self here and I am going to rewrite the signature of those methods to avoid passing in input references (let the return values be return values) so the clear shouldn't be necessary anymore (sorry... I saw this part of my old code while reviewing your CR and I shivered).

GMercat: Ok.

It's never easy to re-read our old code 😄

etr:
Merging. Thanks for the work!

PR #144, opened 2019-01-29
Add private data support for deferred_response

coin0:
This is a PR for additional private data support for deferred_response, it is necessary to carry private information so that cycle_callback could tell different HTTP connections by the customized data. cleanup_callback is used to clean up private data allocated by users.

### DIFF
--- a/examples/minimal_deferred.cpp
+++ b/examples/minimal_deferred.cpp
@@ -18,32 +18,65 @@
      USA
 */
 
+#include <atomic>
 #include <httpserver.hpp>
 
 using namespace httpserver;
 
-static int counter = 0;
+std::atomic<int> reqid;
 
-ssize_t test_callback (char* buf, size_t max) {
-    if (counter == 2) {
+typedef struct {
+    int reqid;
+} connection;
+
+ssize_t test_callback (void* data, char* buf, size_t max) {
+    int reqid;
+    if (data == nullptr) {
+        reqid = -1;
+    } else {
+        reqid = static_cast<connection*>(data)->reqid;
+    }
+
+    // only first 5 connections can be established
+    if (reqid >= 5) {
         return -1;
     }
-    else {
-        memset(buf, 0, max);
-        strcat(buf, " test ");
-        counter++;
-        return std::string(buf).size();
+
+    // respond corresponding request IDs to the clients
+    std::string str = "";
+    str += std::to_string(reqid) + " ";
+    memset(buf, 0, max);
+    std::copy(str.begin(), str.end(), buf);
+
+    // keep sending reqid
+    sleep(1);
+    return (ssize_t)max;
+}
+
+void test_cleanup (void** data)
+{
+    if (*data != nullptr) {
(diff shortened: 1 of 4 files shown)


src/httpserver/deferred_response.hpp
```diff
@@ -35,26 +35,33 @@ namespace details
     ssize_t cb(void*, uint64_t, char*, size_t);
 };
 
-typedef ssize_t(*cycle_callback_ptr)(char*, size_t);
+typedef ssize_t(*cycle_callback_ptr)(void*, char*, size_t);
```
etr: When I was thinking about this feature, I tried to figure how we could avoid the "void*" for the type.

To be honest this is probably one of those cases where "seeing it after it is written might help".

I was thinking we could use a template type to define the first parameter and a "shared_ptr" to pass it through. This would save us from having to use the "void*" and, ideally, from having the cleanup function entirely.

To be honest, I would convert the currently existing "char*" into a shared_ptr if we are perturbing the interface.

test/integ/deferred.cpp
```diff
@@ -67,7 +67,8 @@ class deferred_resource : public http_resource
     public:
         const shared_ptr<http_response> render_GET(const http_request& req)
         {
-            return shared_ptr<deferred_response>(new deferred_response(test_callback, "cycle callback response"));
+            return shared_ptr<deferred_response>(new deferred_response(test_callback, nullptr, nullptr,
```
etr: We should have a separate resource and a separate that is used to verify that the new functionality is working fine.

examples/minimal_deferred.cpp
```diff
@@ -18,32 +18,65 @@
...
 #include <httpserver.hpp>
 
 using namespace httpserver;
 
-static int counter = 0;
+std::atomic<int> reqid;
 
-ssize_t test_callback (char* buf, size_t max) {
-    if (counter == 2) {
+typedef struct {
```
etr: No need for typedefs - we can just have a named struct

examples/minimal_deferred.cpp
```diff
@@ -18,32 +18,65 @@
...
     }
+    data = nullptr;
 }
 
 class deferred_resource : public http_resource {
     public:
         const std::shared_ptr<http_response> render_GET(const http_request& req) {
-            return std::shared_ptr<deferred_response>(new deferred_response(test_callback, "cycle callback response"));
+            // private data of new connections
+            auto priv_data = new connection();
```
etr: Please, refrain from using "auto" for now

etr:
Hey,

thanks for your first contribution.

I left you some comments on the code - hopefully, we can work out a good solution for this (it is something I was looking at given a previous pull request).

Besides the comments, can you please follow one of the templates for the pull request?

etr:
I have an implementation of this that I completed over the weekend (see: https://github.com/etr/libhttpserver/pull/145 ) - let me know if there is still something I am missing from your use-case.

Issue #156, opened 2019-08-14
[BUG] deferred_with_accumulator example using *nix sleep function, incompatible with Windows


deanmsands3:
### Prerequisites

* [X ] Put an X between the brackets on this line if you have checked that your issue isn't already filed: https://github.com/search?l=&q=repo%3Aetr%2Flibhttpserver&type=Issues

### Description

Example deferred_with_accumulator using *nix sleep function, incompatible with Windows

### Steps to Reproduce

1. (Fix the ws2_32 linker error for MinGW)
2. Compile examples
3. Compilation fails on deferred_with_accumulator example

**Expected behavior:**
Proper compilation of all examples

**Actual behavior:**
Compilation fails with deferred_with_accumulator example

**Reproduces how often:**
%100 on any system lacking the C sleep() function

### Versions

* Windows 10x64
* libhttpserver version: git commit 3ffda7b (compiled)
* libmicrohttpd version (please specify whether compiled or packaged)

If you have problems during build:
* Compiler version: MSYS2/MinGW64 GCC 9.2.0
* autotools version: automake 1.16.1, autoconf 2.69

PR #157, opened 2019-08-14
Adapted sleep function in 'examples/deferred_with_accumulator.cpp' for non-*Nix

deanmsands3:
### Identify the Bug

https://github.com/etr/libhttpserver/issues/156

### Description of the Change

Modified deferred_with_accumulator.cpp example
Added thread, chrono headers. Changed sleep() function to C++11

### Alternate Designs

#ifdef Windows hack: https://stackoverflow.com/a/10918411/234826

### Possible Drawbacks

Might add infinitesimal overhead for Unix systems compared to using straight sleep()

### Verification Process

Checked on MinGW64 and Ubuntu x64

### Release Notes

Replaced sleep function with C++11 equivalent.

### DIFF
--- a/examples/deferred_with_accumulator.cpp
+++ b/examples/deferred_with_accumulator.cpp
@@ -18,6 +18,8 @@
      USA
 */
 
+#include <chrono>
+#include <thread>
 #include <atomic>
 #include <httpserver.hpp>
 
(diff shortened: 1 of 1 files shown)


etr:
Thanks for fixing.

Issue #74, opened 2015-04-26
Cannot create shared version of Libhttpserver for MinGW using MSYS


eaguilerai:
I successfully configure libhttpserver on MSYS, using this command:

../configure --prefix=/mingw --build=x86_64-w64-mingw32 --libdir=/mingw/lib64

where /mingw is a mount point to the root directory of MinGW-w64.

Howerver, when I run make, I get the following warnings that do not allow to build the shared versión of libhttpserver:

**\* Warning: linker path does not have real file for library -lregex.
**\* I have the capability to make that library automatically link in when
**\* you link to this library.  But I can only do this if you have a
**\* shared version of the library, which you do not appear to have
**\* because I did check the linker path looking for a file starting
**\* with libregex and none of the candidates passed a file format test
**\* using a file magic. Last file checked: C:/mingw/msys/1.0/opt/windows_64/lib64/libregex.a

**\* Warning: linker path does not have real file for library -lgpg-error.
**\* I have the capability to make that library automatically link in when
**\* you link to this library.  But I can only do this if you have a
**\* shared version of the library, which you do not appear to have
**\* because I did check the linker path looking for a file starting
**\* with libgpg-error and none of the candidates passed a file format test
**\* using a file magic. Last file checked: C:/mingw/msys/1.0/opt/windows_64/lib64/libgpg-error.dll.a

**\* Warning: linker path does not have real file for library -lws2_32.
**\* I have the capability to make that library automatically link in when
**\* you link to this library.  But I can only do this if you have a
**\* shared version of the library, which you do not appear to have
**\* because I did check the linker path looking for a file starting
**\* with libws2_32 and none of the candidates passed a file format test
**\* using a file magic. Last file checked: c:/mingw/msys/1.0/opt/windows_64/mingw/lib/libws2_32.a
**\* The inter-library dependencies that have been dropped here will be
**\* automatically added whenever a program is linked with this library
**\* or is declared to -dlopen it.

**\* Since this library must not contain undefined symbols,
**\* because either the platform does not support them or
**\* it was explicitly requested with -no-undefined,
**\* libtool will only create a static version of it.

I am not sure, but I think -lregex should be -lgnurx instead, which I have it installed in /mingw/bin. As for gpg-error, I also have the shared versión installed in /mingw/bin. And as for ws2_32, I couldn't find any shared versión installed.

deanmsands3:
[...]

PR #158, opened 2019-08-15
Added Winsock2 library for mingw; Not tested for cygwin

deanmsands3:
### Identify the Bug

https://github.com/etr/libhttpserver/issues/74

### Description of the Change

Added NETWORK_LIBS variable in **configure.ac** at CASE **$host**.
Added NETWORK_LIBS="-lws2_32" to mingw CASE
Added $NETWORK_LIBS to LDFLAGS before $REGEX_LIBS for both compile/cross-compile

### Alternate Designs

Adding -lws2_32 to Cygwin CYGWIN, but I haven't confirmed it needs it.
Adding -lws2_32 to REGEX_LIBS but that felt dishonest even though it's the easier option.

### Possible Drawbacks

One MORE variable in configure.ac to keep track of.

### Verification Process

Tested on MSYS/MinGW64, Ubuntu 19.04x64
Initial tests on Cygwin are failing on WinSock2 header test, but I may have "corrupted" my install by installing the mingw64-x86_64-gcc toolchain in Cygwin.

### Release Notes
Fixes the linking failure in Issue 74 for MinGW targets.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -60,10 +60,12 @@ if test x"$samedirectory" = x"no"; then
 fi
 
 is_windows=yes;
+NETWORK_LIBS=""
 case "$host" in
     *-mingw*)
         NETWORK_HEADER="winsock2.h"
         REGEX_LIBS="-lregex -lpthread -no-undefined"
+        NETWORK_LIBS="-lws2_32"
         native_srcdir=$(cd $srcdir; pwd -W)
         ;;
     *-cygwin*)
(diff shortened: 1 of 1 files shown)


etr:
Hey, thanks for sending this over.
I am not sure of what you mean with your comment about cygwin. Are you seeing a change in behavior from before and after your change?
(I might want to test it on cygwin before merging)

deanmsands3:
It doesn't make the Cygwin build process any worse. I just wasn't sure that Cygwin didn't also need it. I'm MSYS/MinGW all the way. Cygwin is just something I dabble in.
Anyway, **configure** is failing on the Winsock2 header test, but I'm not sure it was ever going to work in the first place. If I'm reading things right, Cygwin should not be using Winsock2 at all - even though it has the header.

etr:
sounds fine with what we have. I verified that it doesn't make a difference on Cygwin (will need to fix that separately). Thanks again for your contribution.

Issue #177, opened 2020-02-17
[BUG] General code style issues


bcsgh:
In digging thought the code while debugging my usage of it, I've noticed a number of oddities that seem like things that could easily be cleaned up to improve the code readability.

Would PR's doing general cleaning be welcomed and actively supported?

In general, I think the sort of cleanup described below would improve the code by both making it more readable as well as by making it shorter (in a few cases, by a lot), but it would also be rather invasive in that it would likely touch a significant faction of the code which could cause merge conflicts and the like. That said, most everything I've see would be highly local and simple so any merge conflicts would likely be relatively trivial.

Examples:
-

- Inconsistent formatting:

The formatting style is rather non-uniform (e.g. `if()` and `if ()` are both used, function argument sometimes skip spaces after the comma, most lines are under 80col but a few hundred are not, etc.). With tools like clang-format, it should be a rather mechanical task to identify the most prevalent patterns and apply them globally.

- Pre/post C++11 idioms:

The code uses post C++11 features like `std::move()` but also makes extensive uses of patterns rendered obsolete by C++11, in particular; iterator for loops rather than range for.

- Others:
  - redundant lookups: 
    `if (m.count(k)) return m[k];` vs
    `auto i = m.get(); if (i != m.end()) return i->second;`)
  - redundant constructor definitions (e.g. where `= default` would have the same effect)
  - redundant actions in destructor (clearing a `std::map` that is about to it self be destructed).
  - etc.

These are just the ones I've seen so far and I suspect if I dig, I'd find more:

etr:
Hey,
thanks for offering to help. The lib is showing signs of its age and the stratification of refactors (e.g. introducing C++11, etc.) is glaring.

I'd be happy to review and merge any PR that addresses any of the stylistic issues you mentioned above.

Closing this issue as I guess was more of a general question - feel free to re-open if I haven't fully addressed your ask.

bcsgh:
Part of the reason I opened this bug was as a persisted public place to discuss things. In digging a bit more, it seems that Gitter is what you use for that? However 1) that appears to be yet another chat app with 2) yet another account and 3) it has no activity in the last 4 months ... so I'm not going to use it.

Do you have a suggestion of a better place to discuss things? A mailing list (like what google groups provides) would be ideal.

In the meantime, a few questions:

- Is there a reason that http_response::content isn't included in the copy/move constructor/assignment operators?

[...]

PR #179, opened 2020-02-24
Use =default for constructors and operator= where possible

bcsgh:
### Identify the Bug
etr/libhttpserver#177 (General code style issues)

### Description of the Change
Use `=default` for constructors and `operator=` where possible.

For default construction, this moves the default values to a default member initializer:
https://en.cppreference.com/w/cpp/language/data_members#Member_initialization

This significantly reduce the size of code in a number of places. It also avoids the need to ensure that the list of copied/moved fields matches the list of existing fields (in some case it did not).

### Possible Drawbacks
This removes the "effectively unused" protected member http_response::content. If any client code references that, this will break it.

### Verification Process

    ./bootstrap 
    mkdir build
    cd build
    CPPFLAGS=-I${libmicrohttpd_include} LDFLAGS=-L${libmicrohttpd_libs} ../configure
    make

### Release Notes

Use `=default` for constructors and `operator=` where possible.

### DIFF
--- a/src/httpserver/create_webserver.hpp
+++ b/src/httpserver/create_webserver.hpp
@@ -25,6 +25,7 @@
 #ifndef _CREATE_WEBSERVER_HPP_
 #define _CREATE_WEBSERVER_HPP_
 
+#include <limits>
 #include <stdlib.h>
 #include "httpserver/http_utils.hpp"
 #include "httpserver/http_response.hpp"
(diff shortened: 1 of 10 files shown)


bcsgh:
Apologies for fixing the noexcept issue piecemeal; I have yet to figure out how to reproduce it in a local build. (What I would give for everything to use a *fully* hermetic build system?!)

src/httpserver/http_response.hpp
```diff
@@ -183,14 +144,13 @@ class http_response
...
-        std::string content;
-        int response_code;
+        int response_code = -1;
 
         std::map<std::string, std::string, http::header_comparator> headers;
         std::map<std::string, std::string, http::header_comparator> footers;
         std::map<std::string, std::string, http::header_comparator> cookies;
 
-    	friend std::ostream &operator<< (std::ostream &os, const http_response &r);
+      	friend std::ostream &operator<< (std::ostream &os, const http_response &r);
```
etr: the indentation must have accidentally changed here. Can we fix it?

bcsgh:
### Note
Given that this PR consist mostly of removing code, I suspect the coverage numbers are not meaningful. Also, the delta is of a nature where most possibilities for failure are compile time issues and as such will be detected regardless of if the tests cases cover them.

src/httpserver/create_webserver.hpp
```diff
@@ -500,44 +252,44 @@ class create_webserver
...
-        render_ptr _internal_error_resource;
+        uint16_t _port = DEFAULT_WS_PORT;
+        http::http_utils::start_method_T _start_method = http::http_utils::INTERNAL_SELECT;
+        int _max_threads = 0;
+        int _max_connections = 0;
+        int _memory_limit = 0;
+        size_t _content_size_limit = std::numeric_limits<size_t>::max();
+        int _connection_timeout = DEFAULT_WS_TIMEOUT;
+        int _per_IP_connection_limit = 0;
+        log_access_ptr _log_access = nullptr;
```
etr: As there are many comparisons of these values to 0x0, I'd isolate the change to nullptr from this change. At least we keep the change dedicated to only moving to default constructor.

Also, despite nullptr is 0x0 under the hood, but formally I'd rather have that consistent before changing it as it should act as an abstraction (so I'd rather not assume it is).

src/httpserver/create_webserver.hpp
```diff
@@ -500,44 +252,44 @@ class create_webserver
...
-        bool _tcp_nodelay;
-        render_ptr _not_found_resource;
-        render_ptr _method_not_allowed_resource;
-        render_ptr _internal_error_resource;
+        uint16_t _port = DEFAULT_WS_PORT;
+        http::http_utils::start_method_T _start_method = http::http_utils::INTERNAL_SELECT;
+        int _max_threads = 0;
+        int _max_connections = 0;
+        int _memory_limit = 0;
+        size_t _content_size_limit = std::numeric_limits<size_t>::max();
```
etr: The change from 0 to max is definitely a correct change, but I'd rather do it in a different CR to push before pushing this.

bcsgh: Mind if I do it after this one?

(Git seems to be gratuitously obtuse when it comes to dealing with merges and PRs.)

src/httpserver/http_response.hpp
bcsgh: Reverted.

src/httpserver/create_webserver.hpp
bcsgh: Reverted.

etr:
The change looks good. There have been some uncorrelated failures on Travis (they were having problems at connecting to PPA). I am restarting those to make sure nothing happens there and approving this change.

etr:
Merged. Thanks for all the work and support.

PR #180, opened 2020-02-26
Rename some variable so that all `this->` can be dropped, and do so.

bcsgh:
### Identify the Bug
etr/libhttpserver#177 (General code style issues)

### Description of the Change

Rename some variable so that all `this->` can be dropped, and do so.

`this->` is almost always redundant.  

Side Note: the cases where this required changing the name of local variable to not conflict with member variable is the reason that many style guides (e.g. https://google.github.io/styleguide/cppguide.html#Variable_Names) require different forms for those types of variables. It may be worth renaming all private member variables at some point.

### Alternate Designs

n/a

### Possible Drawbacks

none really.

### Verification Process

Normal build process. CI build to follow

### Release Notes

- Code cleanup relating to `this->`.

### DIFF
--- a/src/http_request.cpp
+++ b/src/http_request.cpp
@@ -37,9 +37,9 @@ struct arguments_accumulator
     std::map<std::string, std::string, http::arg_comparator>* arguments;
 };
 
-void http_request::set_method(const std::string& method)
+void http_request::set_method(const std::string& method_src)
 {
-    this->method = string_utilities::to_upper_copy(method);
+    method = string_utilities::to_upper_copy(method_src);
 }
 
 bool http_request::check_digest_auth(
@@ -76,7 +76,7 @@ bool http_request::check_digest_auth(
 const std::string http_request::get_connection_value(const std::string& key, enum MHD_ValueKind kind) const
 {
     const char* header_c = MHD_lookup_connection_value(
-        this->underlying_connection,
+        underlying_connection,
         kind,
         key.c_str()
     );
@@ -103,7 +103,7 @@ const std::map<std::string, std::string, http::header_comparator> http_request::
     std::map<std::string, std::string, http::header_comparator> headers;
 
     MHD_get_connection_values(
-        this->underlying_connection,
+        underlying_connection,
         kind,
         &build_request_header,
         (void*) &headers
@@ -144,9 +144,9 @@ const std::map<std::string, std::string, http::header_comparator> http_request::
 
 const std::string http_request::get_arg(const std::string& key) const
 {
-    std::map<std::string, std::string>::const_iterator it = this->args.find(key);
+    std::map<std::string, std::string>::const_iterator it = args.find(key);
 
-    if(it != this->args.end())
+    if(it != args.end())
     {
         return it->second;
     }
@@ -157,14 +157,14 @@ const std::string http_request::get_arg(const std::string& key) const
 const std::map<std::string, std::string, http::arg_comparator> http_request::get_args() const
 {
     std::map<std::string, std::string, http::arg_comparator> arguments;
-    arguments.insert(this->args.begin(), this->args.end());
+    arguments.insert(args.begin(), args.end());
 
     arguments_accumulator aa;
-    aa.unescaper = this->unescaper;
+    aa.unescaper = unescaper;
     aa.arguments = &arguments;
 
     MHD_get_connection_values(
-        this->underlying_connection,
+        underlying_connection,
         MHD_GET_ARGUMENT_KIND,
         &build_request_args,
         (void*) &aa
@@ -178,7 +178,7 @@ const std::string http_request::get_querystring() const
     std::string querystring = "";
 
     MHD_get_connection_values(
-        this->underlying_connection,
+        underlying_connection,
         MHD_GET_ARGUMENT_KIND,
         &build_request_querystring,
         (void*) &querystring
--- a/src/http_request.cpp
+++ b/src/http_request.cpp
@@ -37,9 +37,9 @@ struct arguments_accumulator
     std::map<std::string, std::string, http::arg_comparator>* arguments;
 };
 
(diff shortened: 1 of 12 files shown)


bcsgh:
I think the failing coverage is another case of pre-existing conditions.

src/http_request.cpp
```diff
@@ -37,9 +37,9 @@ struct arguments_accumulator
     std::map<std::string, std::string, http::arg_comparator>* arguments;
 };
 
-void http_request::set_method(const std::string& method)
+void http_request::set_method(const std::string& method_src)
```
etr: I'd leave this specific one as is given that it is a setter method

src/httpserver/http_request.hpp
```diff
@@ -275,49 +275,48 @@ class http_request
...
 
         /**
          * Method used to set the content of the request
          * @param content The content to set.
         **/
-        void set_content(const std::string& content)
+        void set_content(const std::string& content_src)
         {
-            this->content = content.substr(0,content_size_limit);
+            content = content_src.substr(0,content_size_limit);
```
etr: as for the other setter I commented, I'd leave this one with "this" that I find more expressive in this case

src/httpserver/http_request.hpp
```diff
@@ -275,49 +275,48 @@ class http_request
...
 
         /**
          * Method used to set the maximum size of the content
          * @param content_size_limit The limit on the maximum size of the content and arg's.
         **/
-        void set_content_size_limit(size_t content_size_limit)
+        void set_content_size_limit(size_t content_size_limit_src)
         {
-            this->content_size_limit = content_size_limit;
+            content_size_limit = content_size_limit_src;
```
etr: same

src/httpserver/http_request.hpp
```diff
@@ -275,49 +275,48 @@ class http_request
...
 
         /**
          * Method used to set the path requested.
          * @param path The path searched by the request.
         **/
-        void set_path(const std::string& path)
+        void set_path(const std::string& path_src)
         {
-            this->path = path;
+            path = path_src;
```
etr: same

src/httpserver/http_request.hpp
```diff
@@ -330,20 +329,20 @@ class http_request
          * Method used to set the request http version (ie http 1.1)
          * @param version The version to set in form of string
         **/
-        void set_version(const std::string& version)
+        void set_version(const std::string& version_src)
         {
-            this->version = version;
+            version = version_src;
```
etr: same

src/httpserver/http_request.hpp
```diff
@@ -330,20 +329,20 @@ class http_request
...
         }
 
         /**
          * Method used to set all arguments of the request.
          * @param args The args key-value map to set for the request.
         **/
-        void set_args(const std::map<std::string, std::string>& args)
+        void set_args(const std::map<std::string, std::string>& args_src)
```
etr: same

test/integ/basic.cpp
```diff
@@ -245,9 +245,9 @@ class error_resource : public http_resource
 class print_request_resource : public http_resource
 {
     public:
-        print_request_resource(std::stringstream* ss)
+        print_request_resource(std::stringstream* ss_src)
         {
-            this->ss = ss;
+            ss = ss_src;
```
etr: same

test/integ/basic.cpp
```diff
@@ -263,9 +263,9 @@ class print_request_resource : public http_resource
 class print_response_resource : public http_resource
 {
     public:
-        print_response_resource(std::stringstream* ss)
+        print_response_resource(std::stringstream* ss_src)
         {
-            this->ss = ss;
+            ss = ss_src;
```
etr: same

bcsgh:
Regarding the setters; I fully agree with your intent, but I strongly *disagree* with your proposed means.

The intent of making it clear where a variable is coming from the object vs. somewhere else is rather desirable, but not just in setters; it is my opinion that that should be done universally.

As noted in the PR description, I'd suggest accomplishing that intent (universally) by making the form of member variable names distinct from locals. The `create_webserver` class already does this by prepending a `_` to it's members and I'd suggest there is more general utility in applying that pattern[1] globally than in adding `this->` on a case-by-cases basis.

If that is going to be the done, then making that transformation as a follow-on PR would be the preferred next step[2] and dropping these `this->` now would be the proper action here.

All that said, that's *my* opinion, and this is *your* code. So I'll just make my case and let you decide.

[1]: I personally prefer trailing `_` as is the google style, but that's even more a matter of taste than tabs vs. spaces.
[2]: that is rendered somewhat more complex by the presence of protected members variables that may be used by sub-classes outside this repo. However, that (having protected data members) is a bit of technical debt that is likely better fixed now than later (https://ericlippert.com/2020/02/27/hundred-year-mistakes/). If you are interested, I could probably dig up a case for why member variable should basically never be protected (tl;dr; something along the lines of don't mix API and implementation).

etr:
I appreciate the feedback but I'd stay with setters parameters matching the object internal one and "this" used. The general rule of thumb is that "this" is /almost/ always redundant. The setters are where /almost/ has value for me in that sentence. I agree with the rest of your CR though, as elsewhere, it is redundant.

create_webserver has been managed differently to allow for members to be named to match those in the webserver class.

On the protected members, sadly that has been poisoning this repo for a long time. Admittedly, they need better handling that I didn't have the time to do. Some of these things have been in the repo for quite a while as the project itself has been around for ~10 and has accumulated tech debt.

etr:
Additional note that I think wasn't clear at all from my previous comment. I would be totally happy with a subsequent change to enforce the use of "_" in front of all class members (thus making "this" redundant for setters). In the context of this PR, I rather prefer having setters using "this" than having parameters with different names.

etr:
Everything merged. Thanks again for your support to the library.

PR #183, opened 2020-03-06
Make #include for internal headers always use fully qualified paths

bcsgh:
### Identify the Bug

#177 (General code style issues)

### Description of the Change
Make #include for internal headers always use fully qualified paths.

This make the code more uniform, removes some question as to what is being refered to and potentially simplifies the needed include search path.

My intent it so follow this up with sorting the includes (I'd suggest https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes) and then adding missing includes (i.e. https://include-what-you-use.org/). Getting things uniform first will also simplify those steps.

### Possible Drawbacks
None?

### Verification Process

./bootstrap
mkdir build ; cs build
../configure
make

### Release Notes
Make #include for internal headers always use fully qualified paths

### DIFF
--- a/src/basic_auth_fail_response.cpp
+++ b/src/basic_auth_fail_response.cpp
@@ -18,7 +18,7 @@
      USA
 */
 
-#include "basic_auth_fail_response.hpp"
+#include "httpserver/basic_auth_fail_response.hpp"
 
 using namespace std;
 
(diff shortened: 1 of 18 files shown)


etr:
Merged. Thanks again for your contributions!

Issue #186, opened 2020-03-08
[BUG] Travic-CI config doesn't build/test for Windows.


bcsgh:
### Prerequisites

* [x] Put an X between the brackets on this line if you have checked that your issue isn't already filed: https://github.com/search?l=&q=repo%3Aetr%2Flibhttpserver&type=Issues

### Description

The current runs on travis-ci.org don't run any Windows cases. However the config does mention windows:
https://github.com/etr/libhttpserver/blob/master/.travis.yml

If I had to guess (and I've never configured Travis so it would be just a guess), I'd suspect that this line here shouldn't have the leading `-`: https://github.com/etr/libhttpserver/blob/master/.travis.yml#L159

### Steps to Reproduce

1. create a PR
2. check the CI link

**Expected behavior:** builds/tests for Windows

**Actual behavior:** Nothing for windows.

**Reproduces how often:** 100%

### Versions
master

etr:
This is not a bug, or at least not in the library. As you can see here: https://github.com/etr/libhttpserver/blob/master/.travis.yml#L123 it is an explicit choice not to run those tests on windows. This is due to the clunky installation of mingw and msys that travis currently has.
An alternative (appveyor) is being evaluated to run those on windows too.

For now, through a quite complex work on the travis configuration, we can automatically run builds on windows that are capturing the majority of the problems that have been appearing on that platform over the years.

For now, until travis doesn't fix their windows support (they are working on it), there is nothing to do here.

bcsgh:
So regardless of if the tests are built, they are not run?

As for the rest, I'm not following; Are you saying that the CI:
- "is building on windows" (I'm seeing evidence that it's not.)
- "could be made to build on windows" (Yes please!)
- "is currently too complex to make build on windows" (Oh, okay.)

Or maybe I'm assuming too much; is travis-ci unable to build a test but then not run it?

etr:
Two different discussions.

1) [Known] Even if building the tests, due to the way they have configured the environment, it will not be able to run them (or, at least, without severe assumptions to how the paths are configured on travis machines, that will break every time they make a change on their side). Because of this, till they are out of their beta on windows, this is not a place where I will be willing to invest.
2) Which I just noticed after my comment. For some reason, with no change to the config it stopped building on windows entirely ( this being the last that executed them https://travis-ci.org/etr/libhttpserver/builds/610168646 ).

Re-opening the issue (and likely will have to open one for travis themselves) to sort 2 out.

PR #187, opened 2020-03-08
Force windows build

etr:
### Identify the Bug

Recovers the build on windows in travis - travis is not building on windows anymore seemingly due to an inversion in priority of exclude and include.

### Description of the Change

As travis seems to have inverted the priority of exclude and include, the windows build is missing. This attempts to obtain the windows build again by:
1) Removing windows from the matrix entirely thus preventing it from building the full set of combinations.
2) Removing the exclude that now has priority over the include
3) Leaving the include (it should allow it even if removed from the matrix).

### Alternate Designs

Full support for the windows build matrix. Not currently possible due to the limited support from travis for windows.

### Possible Drawbacks

Errors in build. Discoverable through CI.

### Verification Process

Running the CI process.

### Release Notes

CI builds on windows again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,18 @@ env:
  - DEBUG="debug" COVERAGE="coverage"
  - DEBUG="nodebug" COVERAGE="nocoverage"
  - LINKING="static"
+before_cache:
+- |-
+    case $TRAVIS_OS_NAME in
+      windows)
+        # https://unix.stackexchange.com/a/137322/107554
+        $msys2 pacman --sync --clean --noconfirm
+        ;;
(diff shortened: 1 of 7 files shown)


etr:
See conversation here too: https://github.com/etr/libhttpserver/issues/186

etr:
The state of build on windows in travis seems to have actually worsened.

I see a large number of timeouts during the installation of msys2 and autotools through choco and pacman.

etr:
For some reason, the re-ordering seemed to have been preventing the import of the correct libraries. Need to investigate - I will use this branch to fix it before pushing out.

PR #199, opened 2020-07-05
Compatibility with libmicrohttpd 0.9.71

bkuhls:
From the libmicrohttpd 0.9.71 release notes:

```
Furthermore, the release introduces an 'enum MHD_Result' instead of
#defines for MHD_YES/MHD_NO. This is intended to make it easier to check
for certain API misuse bugs by providing better types (not everything is
an 'int').  While this does NOT change the binary API, this change
_will_ cause compiler warnings for all legacy code -- until 'int' is
replaced with 'enum MHD_Result'
```

### Identify the Bug

https://github.com/etr/libhttpserver/issues/198

### Description of the Change
Replace 'int' with 'enum MHD_Result'

### Release Notes
Add compatibility with libmicrohttpd 0.9.71

### DIFF
--- a/src/http_request.cpp
+++ b/src/http_request.cpp
@@ -88,7 +88,7 @@ const std::string http_request::get_connection_value(const std::string& key, enu
     return header_c;
 }
 
-int http_request::build_request_header(
+MHD_Result http_request::build_request_header(
         void *cls,
         enum MHD_ValueKind kind,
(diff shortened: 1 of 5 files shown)


etr:
Thanks for the help in keeping the lib compatible with the new version of libmicrohttpd

PR #205, opened 2020-08-08
Update configure and bootstrap, fix compiler warnings in http_endpoin…

beneschtech:

None

### Description of the Change

Updated some of the build scripts to be more portable and to build on FreeBSD.
Fixed a few compiler warnings
Added OS specific readmes for applicable issues to help users

### Possible Drawbacks

Potentially break other platforms

### Verification Process

Built on FreeBSD 12, Fedora 32, CentOS 7, and Mingw64 (Windows 10)

### Release Notes

Added OS specific tips, and cleaned up some compiler warnings

### DIFF
--- a/src/httpserver/details/http_endpoint.hpp
+++ b/src/httpserver/details/http_endpoint.hpp
@@ -120,12 +120,12 @@ class http_endpoint
             return chunk_positions;
         }
 
-        const bool is_family_url() const
+        const bool is_family_url()
         {
             return family_url;
         }
 
-        const bool is_regex_compiled() const
+        const bool is_regex_compiled()
         {
             return reg_compiled;
         }
--- a/src/httpserver/details/http_endpoint.hpp
+++ b/src/httpserver/details/http_endpoint.hpp
@@ -120,12 +120,12 @@ class http_endpoint
             return chunk_positions;
         }
 
-        const bool is_family_url()
+        bool is_family_url() const
         {
             return family_url;
         }
 
-        const bool is_regex_compiled()
+        bool is_regex_compiled() const
         {
             return reg_compiled;
         }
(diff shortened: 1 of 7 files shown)


src/httpserver/details/http_endpoint.hpp
```diff
@@ -120,12 +120,12 @@ class http_endpoint
...
         }
 
-        const bool is_family_url() const
+        const bool is_family_url()
         {
             return family_url;
         }
 
-        const bool is_regex_compiled() const
+        const bool is_regex_compiled()
```
etr, line 128: what was the issue with the is_family_url and is_regex_compiled methods being const?

beneschtech:
Compiler warning. It has const before it still.

etr:
Got it. I think the real issue is the "const bool" (which makes no sense).

Does using "bool is_regex_compiled() const" cause you any warning?

Effectively we want the method to be declared as not modifying the object so I am not sure what the warning against that is.

beneschtech:
As is:
In file included from
../lib/include/httpserver/webserver.hpp:46:../lib/include/httpserver/de
tails/http_endpoint.hpp:123:9: warning: 'const' type qualifier on
return type has no effect [-Wignored-qualifiers]        const bool
is_family_url()
const        ^~~~~~../lib/include/httpserver/details/http_endpoint.hpp:
128:9: warning: 'const' type qualifier on return type has no effect [-
Wignored-qualifiers]        const bool is_regex_compiled() const
Clang complains about the const in front, gcc is okay with either one
removed.
Just pushed new code with const at the end
On Fri, 2020-08-07 at 22:01 -0700, Sebastiano Merlino wrote:
> Got it. I think the real issue is the "const bool" (which makes no
> sense).
> Does using "bool is_regex_compiled() const" cause you any warning?
> Effectively we want the method to be declared as not modifying the
> object so I am not sure what the warning against that is.
> 
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub, or unsubscribe.
-- 
Chris Benesch <chris@beneschtech.com>

etr:
Thanks!
All travis/appveyor tests are passing and all looks good. Thanks a lot for the help.

Issue #202, opened 2020-07-23
[FEATURE] missing support for dualstack ipv6+ipv4


vandalon:
libhttpserver only support either ipv4 OR ipv6.
It would be nice to have support for both at the same time (dualstack).

proposed fix:

webserver.cpp:313;
use MHD_USE_DUAL_STACK instead of MHD_USE_IPv6 or add a new option for dual stack.

(thanks to my colleague for researching this.)

etr:
Interesting. Let me take a look to see if it has other implications within the library with the current implementation (e.g. blocking IPs).

etr:
This feature is now on head. Thanks for requesting it.

PR #206, opened 2020-08-25
Dualstack

etr:

https://github.com/etr/libhttpserver/issues/202

### Description of the Change

Adds an option for dual-stack that allows the servers to use IPV4 and IPV6 simultaneously.

### Alternate Designs

None

### Possible Drawbacks

The testing on dual-stack has been limited. When using dual-stack all IPs are interpreted and returned as IPV6

### Verification Process

travis tests. The main issue has been with Linux on travis that does not support IPV6. Currently we are testing IPV6 features only on OSX and Windows (appveyor) that support IPV6.

### Release Notes

Introduced dual-stack option to allow servers to support IPV4 and IPV6 simultaneously.


etr:
Allowing this to patch - the coverage reduction is a resultant of the tests not running on travis for linux as it has no IPV6 stack. The tests are performed on OSX and windows to prove that the functionality does indeed work. They have also been manually tested on linux.

PR #208, opened 2020-08-27
Travis iwyu

etr:
### Identify the Bug

Include and fwd declarations were inconsistent and often not present when needed (derived from upstream) or redundant.

### Description of the Change

This change adds checks to travis using IWYU (see: https://include-what-you-use.org/ ) preventing this problems from happening transparently. It also fixes the current problems.

### Alternate Designs

N/D. Leave as is.

### Possible Drawbacks

Potential issues on less tested systems (currently no FreeBSD test on travis, nor Cygwin on Windows). Can be covered in the future with more testing through CI/CD.

### Verification Process

Travis runs and Appveyor runs.

### Release Notes

Added checks using IWYU (see: https://include-what-you-use.org/ ) to cleanup includes within the library.

### Notes

The large number of commits is due to travis that forces a push for every test (to verify) and an issue with how clang/iwyu are packetized within ubuntu (which forced a tricky compilation within the travis script).
These changes will be squashed once moved to master.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,7 @@ before_install:
      cmake -G "Unix Makefiles" -DCMAKE_PREFIX_PATH=$CLANG_PREFIX_PATH -DCMAKE_C_COMPILER=$CLANG_BIN_PATH/clang -DCMAKE_CXX_COMPILER=$CLANG_BIN_PATH/clang++ ../include-what-you-use ;
      make ;
      sudo make install ;
+     cd .. ;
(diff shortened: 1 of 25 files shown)


etr:
The change is not applied (currently) to tests and examples.

This PR effectively supersedes: https://github.com/etr/libhttpserver/pull/185 and is related to https://github.com/etr/libhttpserver/issues/177 .

The related PR will be closed accordingly.

PR #209, opened 2020-08-27
Add checks using cpplint

etr:
### Identify the Bug

The code doesn't follow a proper style guide (more or less accumulated over the years).
This results in less readable code for anyone who wants to propose PRs.

### Description of the Change

This change adds a cpplint check on travis and institutes a style guide on the code.
This way, style can be checked on every build (and on every PR). It will also be easier to communicate the style guide to contributors.

### Alternate Designs

Using different tools (e.g. clang-tidy). I've decided to use cpplint because I broadly agree with the underlying style (Google code style) - changes to that are minor and can be managed through configuration.

### Possible Drawbacks

Noisier builds. PRs might fail on contributors because of style errors. This might slow some people down, but the overall benefit in quality is superior. This is also mitigated by the well-known style guide adopted.

### Verification Process

Travis checks.

### Release Notes

Added a cpplint check on travis and institutes a style guide on the code.


etr:
This PR is still in progress (it might take a few commits to complete).

PR #225, opened 2021-03-14
Fix failure on ubuntu 20.04

etr:
Github actions upgraded their ubuntu version and this causes some failures in the CI process.

This PR aims to address those issues.


etr:
See: https://github.com/actions/virtual-environments/issues/1816

PR #218, opened 2021-01-03
Test for Cent / EL 7 and stock compiler and bail if found, known not to work

beneschtech:
Hello there! Welcome. Please follow the steps below to tell us about your contribution.

1. Copy the correct template for your contribution
  - Are you fixing a bug? Copy the template from https://raw.githubusercontent.com/etr/libhttpserver/master/.github/PULL_REQUEST_TEMPLATE/bug_fix.md
  - Are you improving performance? Copy the template https://raw.githubusercontent.com/etr/libhttpserver/master/.github/PULL_REQUEST_TEMPLATE/performance_improvement.md
  - Are you updating documentation? Copy the template from https://raw.githubusercontent.com/etr/libhttpserver/master/.github/PULL_REQUEST_TEMPLATE/documentation.md
  - Are you changing functionality? Copy the template from https://raw.githubusercontent.com/etr/libhttpserver/master/.github/PULL_REQUEST_TEMPLATE/feature_change.md
2. Replace this text with the contents of the template
3. Fill in all sections of the template
4. Click "Create pull request"

PR #227, opened 2021-03-15
Check compiler features

etr:
The library needs a minimal feature-set matching C++14 (e.g. g++ 5.5.0). Some distros still deploy using older versions of gcc and clang. People currently have to check the documentation to verify that their compiler can build - this makes sure that configure checks for that too.


etr:
Ref: https://github.com/etr/libhttpserver/issues/198
Alternative PR: https://github.com/etr/libhttpserver/pull/218

Keeping this one as it should work more generally across OSes and compilers (rather than focusing specifically on rhel and gcc).

Issue #238, opened 2021-11-08
[BUG] library crashes with segfault if render methode returns empty shared_ptr


LeSpocky:
### Prerequisites

* [x] Put an X between the brackets on this line if you have checked that your issue isn't already filed: https://github.com/search?l=&q=repo%3Aetr%2Flibhttpserver&type=Issues

### Description

If class inheriting from http_resource returns a `shared_ptr<http_response>` containing `nullptr` the library crashes with segfault.

### Steps to Reproduce

1. Create a custom resource class inheriting from http_resource
2. Let the render method return a `shared_ptr<http_response>` containing `nullptr`
3. Register an object of your resource to the webserver
4. Call the URL an see the webserver crash

**Expected behavior:** handle the issue gracefully.

**Actual behavior:** segmentation fault.

**Reproduces how often:** 100%

### Versions

* Debian GNU/Linux 10 (buster), amd64
* libhttpserver version 0.18.2-21-gc5cf5ea (master)
* libmicrohttpd version 0.9.62 packaged by Debian

~~If you have problems during build:~~
~~* Compiler version~~
~~* autotools version~~

### Additional Information

None.

PR #239, opened 2021-11-08
Test shared_ptr before accessing members

LeSpocky:
### Identify the Bug

see #238 

### Description of the Change

Add a test case and a fix for the defect described in ticket #238.

### Alternate Designs

This is the most simple fix. shared_ptr object member access happens right after aquiring the shared_ptr and is guarded by a test on .get() now.

### Possible Drawbacks

Minimal (!) perfomance drawback.

### Verification Process

Test suite run.

### Release Notes

Fixed possible segfault if render method returned empty shared pointer.


etr:
Thanks for raising the issue and the PR - it all looks good.

Issue #240, opened 2021-11-09
[BUG] default method-not-allowed response misses HTTP header Allow:


LeSpocky:
### Prerequisites

* [x] Put an X between the brackets on this line if you have checked that your issue isn't already filed: https://github.com/search?l=&q=repo%3Aetr%2Flibhttpserver&type=Issues

### Description

RFC 7231 states in section [6.5.5.  405 Method Not Allowed](https://datatracker.ietf.org/doc/html/rfc7231#section-6.5.5):

> The 405 (Method Not Allowed) status code indicates that the method
> received in the request-line is known by the origin server but not
> supported by the target resource.  The origin server MUST generate an
> Allow header field in a 405 response containing a list of the target
> resource's currently supported methods.

That `Allow:` header is not sent by libhttpserver.

### Steps to Reproduce

1. build with examples
2. run `examples/allowing_disallowing_methods`
3. send POST request, e.g. with `curl -XPOST -v http://localhost:8080/hello`
4. inspect HTTP headers

**Expected behavior:** send that header based on `allowed_methods` member of `class http_resource`

**Actual behavior:** HTTP header `Allow:` is not sent

**Reproduces how often:** always, at least unless webserver is constructed with a custom `method_not_allowed_resource`.

### Versions

* OS version: Debian GNU/Linux 10 (buster)
* libhttpserver version: 0.18.2-22-g263f2bd (current master)
* libmicrohttpd version: 0.9.62 (Debian package)

### Additional Information

curl output (tail):

```
* Connected to localhost (127.0.0.1) port 8080 (#0)
> POST /hello HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.64.0
> Accept: */*
> 
< HTTP/1.1 405 Method Not Allowed
< Connection: Keep-Alive
< Content-Length: 18
< Content-Type: text/plain
< Date: Tue, 09 Nov 2021 13:16:31 GMT
< 
* Connection #0 to host localhost left intact
Method not Allowed
```

Example and explanation for allow header: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Allow

LeSpocky:
[...]

PR #243, opened 2021-11-12
Method not allowed

LeSpocky:
### Identify the Bug

#240 

### Description of the Change

To be able to send along an additional header like `Allow: GET, POST` with an HTTP 405 (Method Not Allowed) response, the `httpserver::http_resource` class had to be extended to expose the allowed methods already stored in that object. Those can be used in the webserver then to build the necessary header line and attach it to the response.

### Alternate Designs

The proposed version was selected based on discussion in the bug report #240.

### Possible Drawbacks

- public API change (`httpserver::http_resource` had to be changed)
- Additional calculation overhead

### Verification Process

Tests were added:

- unit tests for the extension of the class `httpserver::http_resource`
- integration tests for webserver to verify header is actually send in case

### Release Notes

- Class `http_resource` now allows to query the currently set allowed methods
- Webserver now sends 'Allow:' header along with 405 (Method Not Allowed) responses, to comply with HTTP standard (RFC 7231)

### DIFF
--- a/src/webserver.cpp
+++ b/src/webserver.cpp
@@ -601,6 +601,15 @@ MHD_Result webserver::finalize_answer(MHD_Connection* connection, struct details
                 }
             } else {
                 mr->dhrs = method_not_allowed_page(mr);
+
+                vector<string> allowed_methods = hrm->get_allowed_methods();
+                if (allowed_methods.size() > 0) {
+                    string header_value = allowed_methods[0];
+                    for (auto it = allowed_methods.cbegin() + 1; it != allowed_methods.cend(); ++it) {
+                        header_value += ", " + (*it);
+                    }
+                    mr->dhrs->with_header(http_utils::http_header_allow, header_value);
+                }
             }
         } catch(const std::exception& e) {
             mr->dhrs = internal_error_page(mr);
--- a/test/integ/basic.cpp
+++ b/test/integ/basic.cpp
@@ -366,7 +366,7 @@ LT_BEGIN_AUTO_TEST(basic_suite, resource_setting_header)
     curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, writefunc);
     curl_easy_setopt(curl, CURLOPT_WRITEDATA, &s);
     curl_easy_setopt(curl, CURLOPT_HEADERFUNCTION, headerfunc);
-    curl_easy_setopt(curl, CURLOPT_WRITEHEADER, &ss);
+    curl_easy_setopt(curl, CURLOPT_HEADERDATA, &ss);
     res = curl_easy_perform(curl);
     LT_ASSERT_EQ(res, 0);
     LT_CHECK_EQ(s, "OK");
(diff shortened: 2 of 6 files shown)


test/integ/basic.cpp
```diff
@@ -1021,6 +1021,33 @@ LT_BEGIN_AUTO_TEST(basic_suite, url_with_regex_like_pieces)
...
+    curl_easy_setopt(curl, CURLOPT_URL, "localhost:8080/base");
+    curl_easy_setopt(curl, CURLOPT_HTTPGET, 1L);
+    curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, writefunc);
+    curl_easy_setopt(curl, CURLOPT_WRITEDATA, &s);
+    curl_easy_setopt(curl, CURLOPT_HEADERFUNCTION, headerfunc);
+    curl_easy_setopt(curl, CURLOPT_HEADERDATA, &ss);
+    res = curl_easy_perform(curl);
+    LT_ASSERT_EQ(res, 0);
+    long http_code = 0;
+    curl_easy_getinfo(curl, CURLINFO_RESPONSE_CODE, &http_code);
```
LeSpocky, line 1044: cpplint complains about this, however according to https://curl.se/libcurl/c/CURLINFO_RESPONSE_CODE.html *long* is the correct type for `curl_easy_getinfo()` here.

LeSpocky:
The other thing cpplint complains about is header ordering, which will be easy to fix.

test/integ/basic.cpp
LeSpocky: Used `int64_t` like elsewhere in that file now. 🤷🏼‍♂️

LeSpocky:
> The other thing cpplint complains about is header ordering, which will be easy to fix.

Done that and force pushed a new version. Lets see what CI says now?!

src/webserver.cpp
```diff
@@ -601,6 +601,15 @@ MHD_Result webserver::finalize_answer(MHD_Connection* connection, struct details
...
             } else {
                 mr->dhrs = method_not_allowed_page(mr);
+
+                vector<string> allowed_methods = hrm->get_allowed_methods();
+                if (allowed_methods.size() > 0) {
+                    string header_value = allowed_methods[0];
+                    for (auto it = allowed_methods.cbegin() + 1; it != allowed_methods.cend(); ++it) {
+                        header_value += ", " + (*it);
+                    }
+                    mr->dhrs->with_header("Allow", header_value);
```
etr, line 611: Can we use: https://github.com/etr/libhttpserver/blob/263f2bdef4d5f8887a46c19fa08695170ad679f7/src/http_utils.cpp#L135

etr (commented):
It all looks good, only one comment to use the constant defined in http_utils for the "Allow" header.

src/webserver.cpp
LeSpocky: Sure, no problem. Force pushed the updated branch.

Issue #242, opened 2021-11-11
[BUG] Debug print of upload_data does not consider unterminated strings


LeSpocky:
### Prerequisites

* [x] Put an X between the brackets on this line if you have checked that your issue isn't already filed: https://github.com/search?l=&q=repo%3Aetr%2Flibhttpserver&type=Issues

### Description

Debug output fills the console with garbage, if `upload_data` is no terminated string, like this:

```
Writing content: {"login":{"pass":"config","user":"admin"}}��������������������������������������������������
�������������������������������������������������������������������������������������������������������������
�������������������������������������������������������������������������������������������������������������
�������������������������������������������������������������������������������������������������������������
�������������������������������������������������������������������������������������������������������������
�������������������������������������������������������������������������������������������������������������
�������������������������������������������������������������������������������������������������������������
�������������������������������������������������������������������������������������������������������������
�������������������������������������������������������������������������������������������������������������
�������������������������������������������������������������������������������������������������������������
�������������������������������������������������������������������������������������������������������������
�������������������������������������������������������������������������������������������������������������
�������������������������������������������������������������������������������������������������������������
�������������������������������������������������������������������������������������������������������������
�������������������������������������������������������������������������������������������������������������
�������������������������������������������������������������������������������������������������������������
�������������������������������������������������������������������������������������������������������������
�������������������������������������������������������������������������������������������������������������
�������������������������������������������������������������������������������������������������������������
�������������������������������������������������������������������������������������������������������������
�������������������������������������������������������������������������������������������������������������
�������������������������������������������������������������������������������������������������������������
�������������������������������������������������������������������������������������������������������������
�������������������������������������������������������������������������������������������������������������
�������������������������������������������������������������������������������������������������������������
�������������������������������������������������������������������������������������������������������������
�������������������������������������������������������������������������������������������������������������
�������������������������������������������������������������������������������������������������������������
�������������������������������������������������������������������������������������������������������������
�������������������������������������������������������������������������������������������������������������
�������������������������������������������������������������������������������������������������������������
�������������������������������������������������������������������������������������������������������������
�������������������������������������������������������������������������������������������������������������
�������������������������������������������������������������������������������������������������������������
�������������������������������������������������������������������������������������������������������������
�������������������������������������������������������������������������������������������������������������
�������������������������������
```

### Steps to Reproduce

1. Configure libhttpserver with `--enable-debug`
2. Build libhttpserver
3. In your app source create and hook up a resource
[...]

PR #244, opened 2021-11-12
webserver: Fix garbage output on debug print

LeSpocky:
### Identify the Bug

#242 

### Description of the Change

Consider the upload date size passed by libmicrohttpd to construct a std::string, instead of blindly relying on null termination.

### Alternate Designs

N/A

### Possible Drawbacks

*unsure*

### Verification Process

Logging output with that fix:

```
Writing content: {"login":{"pass":"admin","user":"admin"}}
Writing content: {"login":{"pass":"admin","user":"admin"}}
```

(See #242 for how it was before.)

### Release Notes

N/A


etr:
Thanks for taking care of this!

Issue #229, opened 2021-04-10
[BUG] Unable to access Content-Type of file when part of multipart form


jmslocum:
### Prerequisites

* [X ] Put an X between the brackets on this line if you have checked that your issue isn't already filed: https://github.com/search?l=&q=repo%3Aetr%2Flibhttpserver&type=Issues

### Description

When uploading a file as part of a multipart form, the Content-Type, and filename are not accessible from the http_request object. I have tried get_args(), get_headers(), and get_footers(). However if i dump the body with get_content(), I can see the fields in the body, but they are not hoisted into higher level objects. 

### Example Code
```c++
#include <iostream>
#include <algorithm>

#include <httpserver.hpp>

using namespace httpserver;

class file_resource : public http_resource {
  public :
    const std::shared_ptr<http_response> render_POST(const http_request& req) {

      //Loop through the headers map and output them
      const auto& headers = req.get_headers();
      std::for_each(headers.begin(), headers.end(), [](auto& p){
        std::cout << p.first << " -> " << p.second << std::endl;
      });

      //Loop through the args map and output them
      const auto& args = req.get_args(); 
      std::for_each(args.begin(), args.end(), [](auto& p) {
        std::cout << p.first << " -> " << p.second << std::endl;
      });

      std::cout << "Body: " << req.get_content() << std::endl;
      return std::shared_ptr<http_response>(new string_response("OK"));
    }
};

void custom_access_log(const std::string& url) {
  std::cout << "Request to: " <<  url << std::endl;
}

void custom_error_log(const std::string& error) {
  std::cout << "Error: " << error << std::endl;
}

int main() {
  webserver ws = create_webserver(8080)
    .log_access(custom_access_log)
    .log_error(custom_error_log);

  file_resource fr;

  ws.register_resource("/file", &fr);

  ws.start(true);

  return 0;
}
```

### Steps to Reproduce

1. Compile example code above
2. `curl -X POST -F upload=@<some_file>.jpg localhost:8080/file`
3. the "upload" argument will have the file data, but the Content-Type, and original filename are not accessible 
4. Looking at the output, you can see the get_contents() string contains the Content-Type and filename
```
Method: POST, Body: --------------------------ea9b4d3e74b2950b
Content-Disposition: form-data; name="upload"; filename="suite_and_tie.jpg"
Content-Type: image/jpeg
```

**Expected behavior:** 
Some way to access the filename and Content-Type of the file 

**Actual behavior:** 
No way to access it easily

[...]

PR #246, opened 2021-11-17
WIP: Draft new file upload handling

LeSpocky:
### Identify the Bug

#229

### Description of the Change

This is an attempt to address the current shortcomings of big file upload as already discussed in #229.

The idea is to avoid storing possibly huge file upload in memory, at least not by default. The post processor mechanism from libmicrohttpd is used for parsing mime-types *multipart/form-data* and *application/x-www-form-urlencoded* (don't reinvent the wheel!) and text based key/value pairs are put to args as usual. 

Parts with a filename should be written to some C++ stream, a file stream would fit my needs and is shown here as proof of concept. Maybe we want to generalize that somehow? Later?

Initial version of this PR is a draft as base for discussion.

### Alternate Designs

Open for suggestions.

### Possible Drawbacks

It is not possible to get the whole multipart/form-data through `http_request::get_content()` anymore.

### Verification Process

Works on the bench with an example app not shown here. Should be improved of course.

### Release Notes

*tbd*

### DIFF
--- a/src/httpserver/create_webserver.hpp
+++ b/src/httpserver/create_webserver.hpp
@@ -296,6 +296,11 @@ class create_webserver {
          return *this;
      }
 
+     create_webserver& post_upload_dir(const std::string& post_upload_dir) {
+         _post_upload_dir = post_upload_dir;
+         return *this;
+     }
+
      create_webserver& single_resource() {
          _single_resource = true;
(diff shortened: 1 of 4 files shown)


LeSpocky:
#229 was solved by #257, and that's a better solution.

Issue #241, opened 2021-11-09
[BUG] use C++11 keyword nullptr


LeSpocky:
### Prerequisites

* [x] Put an X between the brackets on this line if you have checked that your issue isn't already filed: https://github.com/search?l=&q=repo%3Aetr%2Flibhttpserver&type=Issues

### Description

C++ from C++11 onwards has the keyword **nullptr** to help compiler and developer communicating intent when using null pointers. Use that instead of confusing fellow developers with unusual variants of null.

### Steps to Reproduce

*n.a.*

**Expected behavior:** Use `nullptr` if null pointer is meant.

**Actual behavior:** `0x0` is used in huge parts of the codebase.

**Reproduces how often:** always.

### Versions

* OS version: *n.a.*
* libhttpserver version: 0.18.2-22-g263f2bd (current master)
* libmicrohttpd version: *n.a.*

### Additional Information

According to #227 C++14 is required for libhttpserver already, so using C++11 features should be fine.

LeSpocky:
Somehow related to #177, did not see that one before. 😳

PR #247, opened 2021-11-23
Use keyword nullptr where possible

LeSpocky:
### Identify the Bug

#241 

### Description of the Change

Replace every usage of 0x0 or NULL with nullptr, which is available since C++11.  

`NULL`, or `0x0`, or `nullptr`, or even `0` was used over the entiry code base before for expressing the concept of a null pointer. Use the C++ special word **nullptr** instead to unify the code base and minimize friction with expectations of modern C++ developers.

### Alternate Designs

Using `NULL` instead would have been possible to unify the code base and show developer a null pointer is meant, but that's not C++ and it lacks the advantages of the nullptr keyword.

### Possible Drawbacks

Maybe such style changes are not desired? No functional drawbacks.

### Verification Process

Ran cpplint and unit tests.

### Release Notes

N/A

### DIFF
--- a/examples/service.cpp
+++ b/examples/service.cpp
@@ -158,7 +158,7 @@ int main(int argc, char **argv) {
     while ((c = getopt(argc, argv, "p:k:c:sv?")) != EOF) {
         switch (c) {
         case 'p':
-            port = strtoul(optarg, NULL, 10);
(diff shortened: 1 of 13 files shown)


etr:
Thanks for the great piece of cleanup.

Issue #248, opened 2021-12-03
[BUG] Wrong default MIME type on file_response


LeSpocky:
### Prerequisites

* [x] Put an X between the brackets on this line if you have checked that your issue isn't already filed: https://github.com/search?l=&q=repo%3Aetr%2Flibhttpserver&type=Issues

### Description

When using the constructor of `httpserver::file_response` (derived from `httpserver::http_response`) you can omit the second and third parameter. The content type of the response is set to a default value (*text/plain*) then, which might be harmful or lead to annoying output, if the content of the actual file is unknown, and you can not be sure it is actually text. The default should be set to a MIME type which would match arbitrary text and binary data files: *application/octet-stream*.

### Steps to Reproduce

1. Put some binary file into */path/to/source/examples*
2. Change *examples/minimal_file_response.cpp* like this (filename of that binary file is *Screenshot_20211110_102500.png* in this case):

```diff
diff --git a/examples/minimal_file_response.cpp b/examples/minimal_file_response.cpp
index cc29228..e7f6ae4 100644
--- a/examples/minimal_file_response.cpp
+++ b/examples/minimal_file_response.cpp
@@ -23,7 +23,7 @@
 class file_response_resource : public httpserver::http_resource {
  public:
      const std::shared_ptr<httpserver::http_response> render_GET(const httpserver::http_request&) {
-         return std::shared_ptr<httpserver::file_response>(new httpserver::file_response("test_content", 200, "text/plain"));
+         return std::shared_ptr<httpserver::file_response>(new httpserver::file_response("Screenshot_20211110_102500.png", 200));
      }
 };
 
```

3. build
4. cd into *build/examples*
5. run *./minimal_file_response*
6. open browser
7. open developer tools and network analysis or wireshark
8. open http://localhost:8080/hello
9. inspect output in broswer and response header lines, especially `Content-Type:`

**Expected behavior:** content-type header is recommended default for binary data

**Actual behavior:** content-type header is *text/plain*

**Reproduces how often:** always

### Versions

* OS version: Debian GNU/Linux 10 (buster)
* libhttpserver version: master (0.18.2-26-g0a02a1e)
* libmicrohttpd version: 0.9.62-1 (Debian package)

### Additional Information

Background:

- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types

[...]

PR #249, opened 2021-12-03
file_response: Change default argument for content-type

LeSpocky:
### Identify the Bug

#248 

### Description of the Change

The constructor of class `file_response` can be used with default arguments. Content-type was set to *text/plain* before which is too specific for using the class with unknown files, and could lead to all kinds of broken behavior in different web browsers.

### Alternate Designs

MIME type *application/octet-stream* is what we used for that use-case successfully in different applications. Probably other MIME types starting with *application* would also work, but are not recommended.

### Possible Drawbacks

User might depend on the previous behavior.

### Verification Process

Added a test.

### Release Notes

- Class *file_response* changed default content-type from *text/plain* to *application/octet-stream*


LeSpocky:
Since the change is in the header file, this changes the API, doesn't it?

etr:
Sorry for the delay of the answer - I've been almost afk for the past couple of weeks.

It does indeed change the API. While I tend minimize the number of backward-incompatible interface changes, I think what you are doing here is necessary (and until we are under v1.0, I am not guaranteeing full backward compatibility).

Issue #255, opened 2022-01-27
[BUG] unpredictable results in file_response with filename pointing into the void


LeSpocky:
### Prerequisites

* [x] Put an X between the brackets on this line if you have checked that your issue isn't already filed: https://github.com/search?l=&q=repo%3Aetr%2Flibhttpserver&type=Issues

### Description

When creating an object of class `httpserver::file_response` with a `filename` not related to an actually existing file, the behavior of the application is more or less unpredictable depending on which version of libmicrohttpd is used.  We saw HTTP responses with status 200 and an empty data part, but the app might as well segfault.

### Steps to Reproduce

1. Create an instance of class `httpserver::file_response` with a `filename` pointing to a non existent file
2. Use that object as a response for an actual request

**Expected behavior:** webserver returns a sensible HTTP status code like 500 (Internal Server Error) or better 404 (Not Found).

**Actual behavior:** webserver crashes (segfault), does not answer, or even returns code passed to constructor of `file_response`

**Reproduces how often:** always

### Versions

* OS version
  * Debian GNU/Linux 10 (buster)
  * custom built embedded firmware
* libhttpserver version: master, self compiled
* libmicrohttpd version: multiple self compiled from v0.9.53 up to v0.9.73

If you have problems during build: *build is fine*

### Additional Information

What happens is `httpserver::file_response::get_raw_response()` might return a `nullptr` which is then passed on to `http_response::decorate_response()` which itself calls some libmicrohttpd functions like `MHD_add_response_header()` or `MHD_add_response_footer()` which might or might not crash depending on version.

Will show a unit test reproducing this.

LeSpocky:
See #256 for the promised unit test.

LeSpocky:
FTR: There's an ongoing discussion about how libmicrohttpd should behave in case someone passes them a nullptr: https://lists.gnu.org/archive/html/libmicrohttpd/2022-01/msg00018.html

LeSpocky:
> FTR: There's an ongoing discussion about how libmicrohttpd should behave in case someone passes them a nullptr: https://lists.gnu.org/archive/html/libmicrohttpd/2022-01/msg00018.html

From that discussion it becomes clear libhttpserver should not assume MHD checks if some argument is a nullptr. As a MHD user we must ensure we do not pass nullptr to libmicrohttpd.

etr:
Point taken - I agree with you (and them) that we should check on this side of the wall.

LeSpocky:
[...]

PR #256, opened 2022-01-27
File response

LeSpocky:
### Identify the Bug

See: #255 

### Description of the Change

Adds a unit test highlighting the bug, no real fix so far.

### Alternate Designs

This snippet at least fixes the webserver to return a 500 Internal Server Error:

```diff
diff --git a/src/webserver.cpp b/src/webserver.cpp
index a1392bd..4c4a034 100644
--- a/src/webserver.cpp
+++ b/src/webserver.cpp
@@ -623,6 +623,10 @@ MHD_Result webserver::finalize_answer(MHD_Connection* connection, struct details
     try {
         try {
             raw_response = mr->dhrs->get_raw_response();
+            if (raw_response == nullptr) {
+                mr->dhrs = internal_error_page(mr);
+                raw_response = mr->dhrs->get_raw_response();
+            }
         } catch(const std::invalid_argument& iae) {
             mr->dhrs = not_found_page(mr);
             raw_response = mr->dhrs->get_raw_response();
```

The unit test just fails different then, but that confirms the unit test itself is ok:

```
Running test (25): file_serving_resource_missing
[ASSERT FAILURE] (../../test/integ/basic.cpp:923) - error in "file_serving_resource_missing": 500!=404
- Time spent during "file_serving_resource_missing": 0.668945 ms
```

### Possible Drawbacks

n.a.

### Verification Process

Tested against multiple different version of libmicrohttpd.

### Release Notes

n.a.


LeSpocky:
As discussed in #255 that diff could be a solution already. The unit test could test on HTTP status NOT equal to 200 instead? Or test equal to 500?

LeSpocky:
I force pushed a possible solution as discussed in #255, but did not change description etc. of this PR, yet.  Should I?

src/file_response.cpp
```diff
@@ -31,7 +31,11 @@ namespace httpserver {
 
 MHD_Response* file_response::get_raw_response() {
     int fd = open(filename.c_str(), O_RDONLY);
-    size_t size = lseek(fd, 0, SEEK_END);
+    if (fd == -1) return nullptr;
+
+    off_t size = lseek(fd, 0, SEEK_END);
```
LeSpocky, line 36: Type of `size` taking return value of `lseek()` should be `off_t`. That was introduced wrong with f9b76910. Could be fixed in a separate commit or just fixed here en passant.

LeSpocky:
I just discovered another problem: if that `filename` points to a directory instead of a file … 🙄

LeSpocky:
Those unit test are very repetitive. Maybe that can be improved later?

etr:
> Those unit test are very repetitive. Maybe that can be improved later?

Yeah, I think it would be worth keeping a change to those in an isolated change.

src/httpserver/file_response.hpp
```diff
@@ -37,6 +37,18 @@ class file_response : public http_response {
...
 
+     /**
+      * Constructor of the class file_response.  You usually use this to pass a
+      * filename to the instance.
+      * @param filename Name of the file which content should be sent with the
+      *                 response.  User must make sure file exists and is a
+      *                 regular file, otherwise libhttpserver will return a
+      *                 generic response with HTTP status 500 (Internal Server
+      *                 Error).
+      * @param response_code Optional: HTTP response code in good case.
```
etr, line 48: Worth stating the default is a 200 - HTTP OK.

src/httpserver/file_response.hpp
```diff
@@ -37,6 +37,18 @@ class file_response : public http_response {
...
+      * @param content_type Optional: Mime type of the file content,
```
etr, line 49: Worth stating this defaults to octect/stream

etr (commented):
Can we also add information about the method new failing path in the README file?

LeSpocky:
> Can we also add information about the method new failing path in the README file?

Not sure in what direction that should go?  There is this section *Building responses to requests* which describes the different flavors of *http_response* classes.  The improved failure handling was in the `get_raw_response()` method and over where it is called in `finalize_answer()`.  Instead of crashing or sending back potentially nonsense responses, libhttpserver returns 500 now, and explicitly expects the user to make sure a file used for the `file_response` must exist and be a regular file.  It was a good idea to check that on user side before already.  What do you have in mind should be added to the README and where? 🤔

etr:
In the section called "Building responses to requests" there is piece dedicated to the file_response.

> file_response(const std::string& filename, int response_code = 200, const std::string& content_type = "text/plain"): Uses the filename passed in construction as pointer to a file on disk. The body of the HTTP response will be set using the content of the file. The other two optional parameters are the response_code and the content_type. You can find constant definition for the various response codes within the http_utils library file.

Don't need to make it too complicated. I'd just change it as below (the diff is in bold):

> file_response(const std::string& filename, int response_code = 200, const std::string& content_type = "text/plain"): Uses the filename passed in construction as pointer to a file on disk. The body of the HTTP response will be set using the content of the file. **The file must be a regular file and exist on disk. Otherwise libhttpserver will return an error 500 (Internal Server Error).** The other two optional parameters are the response_code and the content_type. You can find constant definition for the various response codes within the http_utils library file.

PR #257, opened 2022-02-08
Add options to store files directly to the file system and to omit duplicating files in memory

ctomahogh:
https://github.com/etr/libhttpserver/issues/229

### Description of the Change

As discussed in the issue this is an approach to add file handling options for the user of the library
- Option to specify the upload target (memory only, memory and disk, disk only)
- Option to specify the target directory if files are stored to disc (default to `/tmp`)
- Option to let libhttpserver create a random unique filename for stored files rather than using the provided one

Additionally this patch adds an option to avoid extensive memory usage by not duplicating processed data (by the post processor of libmicrohttpd) to the content. Especially when uploading bigger files this increased the memory usage massively.

### Alternate Designs

The option to create a random filename could be enhanced to offer the possibility to sanitize the provided filename rather than creating a unique one

### Possible Drawbacks

Currently if the uploaded file path already exists (if files are stored to disc) the old file gets silently overwritten (will not happen when unique filenames are enabled). This could be enhanced by giving the libhttpserver another option to specify what should happen in this case (like `overwrite_files_if_exist` and `no_overwrite_files_if_exist`)

Otherwise there should not be drawbacks for current users, as the default values for all introduced options preserve the current behavior.

### Verification Process

- Verification by using a single page app which uses a libhttpserver based application as backend which used the previous behavior (now default)
- Adjusting the backend application to use the new options (in this step all possible combinations where tested)

### Release Notes

_tbd_


src/http_utils.cpp
```diff
@@ -193,6 +193,8 @@ const char* http_utils::http_post_encoding_multipart_formdata = MHD_HTTP_POST_EN
 const char* http_utils::application_octet_stream = "application/octet-stream";
 const char* http_utils::text_plain = "text/plain";
 
+const char* http_utils::upload_filename_template = "libhttpserver.XXXXXX";
```
LeSpocky, line 196: Maybe `static const char*` because not used outside of this source file?

src/http_utils.cpp
```diff
@@ -216,6 +218,22 @@ std::string http_utils::standardize_url(const std::string& url) {
     return result;
 }
 
+std::string http_utils::generate_random_upload_filename(std::string &directory) {
+    char *filename = NULL;
```
LeSpocky, line 222: `nullptr` … see #247.

src/http_utils.cpp
```diff
@@ -216,6 +218,22 @@ std::string http_utils::standardize_url(const std::string& url) {
...
+    if (asprintf(&filename, "%s/%s", directory.c_str(), http_utils::upload_filename_template) == -1) {
+        return "";
+    }
```
LeSpocky, line 225: `asprintf()` is a GNU extension, not in C or POSIX. Not sure if this is okay in libhttpserver. 

Also: just returning an empty string on error is somewhat surprising to me. Did not check users of this method. Declaration has no documentation on that.

src/httpserver/http_utils.hpp
```diff
@@ -235,8 +245,12 @@ class http_utils {
      static const char* application_octet_stream;
      static const char* text_plain;
 
+     static const char* upload_filename_template;
+
      static std::vector<std::string> tokenize_url(const std::string&, const char separator = '/');
      static std::string standardize_url(const std::string&);
+
+     static std::string generate_random_upload_filename(std::string &directory);
```
LeSpocky, line 253: cpplint warns here:

    src/httpserver/http_utils.hpp:253:  Is this a non-const reference? If so, make const or use a pointer: std::string &directory  [runtime/references] [2]

LeSpocky (commented):
Thanks for putting so much effort in this topic, for me it looks like you are following a good concept here.  I added some comments from my perspective as a previous contributor of minor fixes, you'll see I'm not aware of some design principles of the library.  Reviewed nevertheless, because I'm highly interested in solving #229 😃 .  Some more things, like those mentioned inline rather nitpicking, than serious flaws:

1. Run `cpplint --extensions=cpp,hpp --headers=hpp --recursive .` before pushing.
2. Would be not my style to put all this in a single commit, because that makes review harder, but @etr squashes them anyways on merging, so this is probably fine.  However the commit message should reflect the reasons for the changes in standard Git style: first line a short (!) summary, followed by an empty line, followed by reasoning for why this change.
3. There are no unit tests for this at all. Yet? 😉 

Again, I think I would want to use your patch eventually, seems reasonable and clear from a library users point of view. 👍🏼

src/http_request.cpp
```diff
@@ -127,6 +127,10 @@ const std::map<std::string, std::string, http::arg_comparator> http_request::get
     return arguments;
 }
 
+file_info_s &http_request::get_or_create_file_info(const char *upload_file_name) {
```
etr, line 130: nit: we generally attach '&' and '*' to the preceding term in declarations of inputs and output (check across the whole CR).

Also, I'd prefer passing around `std::string` over `const char*`.

Can the return be marked const?

src/http_utils.cpp
```diff
@@ -216,6 +218,22 @@ std::string http_utils::standardize_url(const std::string& url) {
     return result;
 }
 
+std::string http_utils::generate_random_upload_filename(std::string &directory) {
```
etr, line 221: Both input and output can be const.

src/http_utils.cpp
```diff
@@ -216,6 +218,22 @@ std::string http_utils::standardize_url(const std::string& url) {
...
+    char *filename = NULL;
+    if (asprintf(&filename, "%s/%s", directory.c_str(), http_utils::upload_filename_template) == -1) {
+        return "";
+    }
+    int fd = mkstemp(filename);
```
etr, line 226: Not sure how portable mkstemp is (to be fair, the autobuild on windows will say that).

I think you might want to use a combination of `tpnam` and `fstream`. See: https://stackoverflow.com/a/7778959/756399

src/httpserver/create_webserver.hpp
```diff
@@ -296,6 +296,36 @@ class create_webserver {
...
+         _put_processed_data_to_content = false;
+         return *this;
+     }
+
+     create_webserver& put_processed_data_to_content() {
+         _put_processed_data_to_content = true;
+         return *this;
+     }
+
+     create_webserver& file_upload_target(file_upload_target_T file_upload_target) {
```
etr, line 309: can we use a const reference?

src/httpserver/http_request.hpp
```diff
@@ -135,6 +135,21 @@ class http_request {
      **/
      const std::map<std::string, std::string, http::arg_comparator> get_args() const;
 
+     /**
+      * Method to get or create a file info struct in the map if the provided filename is already in the map, return the exiting file info struct, otherwise create one in the map and return it.
+      * @param upload_file_name he file name the user uploaded (this is the identifier for the map entry)
+      * @result a file info struct file_info_s
```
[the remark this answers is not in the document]
etr: nit: We sit around at 150 (ideally), but still below this one.

src/httpserver/http_request.hpp
```diff
@@ -135,6 +135,21 @@ class http_request {
...
+     **/
+     file_info_s &get_or_create_file_info(const char *upload_file_name);
```
etr, line 143: I think that the output can be marked as const. The input can use a reference to `std::string`

src/httpserver/http_request.hpp
```diff
@@ -301,6 +317,17 @@ class http_request {
          }
      }
 
+     /**
+      * Method used to set a file info into the map by key.
+      * @param upload_file_name The file name the user uploaded (identifying the map entry)
+      * @param file_system_file_name The path to the file in the file system (the name may not be the original name depending on configuration of the webserver)
+      * @param size The size in number of char of the file.
+     **/
+     void set_file(const char* upload_file_name, const char* file_system_file_name, size_t file_size) {
```
etr, line 326: Let's use strings references on method signatures as much as we can (I am fine with internal conversions, but I'd preserve the interface, if possible).

src/httpserver/webserver.hpp
```diff
@@ -162,6 +162,10 @@ class webserver {
      const bool regex_checking;
      const bool ban_system_enabled;
      const bool post_process_enabled;
+     bool put_processed_data_to_content;
+     file_upload_target_T file_upload_target;
+     std::string post_upload_dir;
```
etr, line 167: Can these two be marked `const`?

src/webserver.cpp
```diff
@@ -454,13 +458,49 @@ MHD_Result webserver::post_iterator(void *cls, enum MHD_ValueKind kind,
...
     std::ignore = content_type;
     std::ignore = transfer_encoding;
     std::ignore = off;
 
     struct details::modded_request* mr = (struct details::modded_request*) cls;
-    mr->dhr->set_arg(key, mr->dhr->get_arg(key) + std::string(data, size));
+
+    if (filename == nullptr || mr->ws->file_upload_target != FILE_UPLOAD_DISK_ONLY) {
+        mr->dhr->set_arg(key, mr->dhr->get_arg(key) + std::string(data, size));
+    }
```
etr, line 469: Would this prevent the library from parsing the other POST parameters whenever the configuation is on?

I guess a unit test would help us see if that's the case.

src/webserver.cpp
```diff
@@ -454,13 +458,49 @@ MHD_Result webserver::post_iterator(void *cls, enum MHD_ValueKind kind,
...
+        // either get the existing file info struct or create a new one in the file map
+        file_info_s &file_info = mr->dhr->get_or_create_file_info(filename);
+        // if the file_system_file_name is not filled yet, this is a new entry and the name has to be set
+        // (either random or copy of the original filename)
+        if (file_info.file_system_file_name.empty()) {
+            if (mr->ws->generate_random_filename_on_upload) {
+                file_info.file_system_file_name = http_utils::generate_random_upload_filename(mr->ws->post_upload_dir);
+            }
+            else {
+                file_info.file_system_file_name = mr->ws->post_upload_dir + "/" + std::string(filename);
```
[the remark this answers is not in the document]
etr: It does (it auto-compiles there on commit).

We might want to just use a preferred separator. The simplest would be to have  a quick inline like this: https://stackoverflow.com/a/12971535/756399

etr (commented):
Overall, thanks a lot for the change; you did a great amount of work here, and I think we are in a pretty good state.

I left you a few comments and tried to avoid overlapping with @LeSpocky.

I think the only major one (besides the inlines) are unit tests that are current missing and the README documentation that is not updated. Let's try to update both, but we are basically near done here.

Note: While I squash on merge, I'd still recommend splitting the changes when you layer them on each other (as it will speed-up the review process).

src/httpserver/http_utils.hpp
```diff
@@ -64,6 +64,16 @@ typedef int MHD_Result;
 
 namespace httpserver {
 
+enum file_upload_target_T{
+     FILE_UPLOAD_MEMORY_ONLY,
+     FILE_UPLOAD_DISK_ONLY,
+     FILE_UPLOAD_MEMORY_AND_DISK,
+};
```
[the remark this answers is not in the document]
ctomahogh: Fixed

src/httpserver/http_resource.hpp
```diff
@@ -47,6 +47,7 @@ void resource_init(std::map<std::string, bool>* res);
 /**
  * Class representing a callable http resource.
 **/
+
```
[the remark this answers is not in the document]
ctomahogh: Accident, as I added some stuff here, and obviously did not fully rollback the change.
Fixed

src/webserver.cpp
```diff
@@ -454,13 +458,49 @@ MHD_Result webserver::post_iterator(void *cls, enum MHD_ValueKind kind,
...
+            }
+            /* to not append to an already existing file, delete an already existing file the file */
```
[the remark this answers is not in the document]
ctomahogh: Fixed

src/webserver.cpp
```diff
@@ -454,13 +458,49 @@ MHD_Result webserver::post_iterator(void *cls, enum MHD_ValueKind kind,
...
+            unlink(file_info.file_system_file_name.c_str());
+        }
+
+        // if multiple files are uploaded, a different filename indicates the start of a new file, so close the previous one
+        if (!mr->upload_filename.empty() && 0 != strcmp(filename, mr->upload_filename.c_str())) {
+            mr->upload_ostrm.close();
+        }
+
+        if (!mr->upload_ostrm.is_open()) {
+            mr->upload_filename = filename;
```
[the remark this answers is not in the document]
ctomahogh: mr->upload_filename is only used internally to keep track of the currently used file

The post_iterator is called multiple times for each file (in chunks of 32 k). The currently used filename and filestream are stored in the modded_request to avoid closing and reopening the very same file for every 32 k (could decrease the upload performance). If the first file is finished and the second is started, mr->upload_filename will be overwritten to the second (or following) file.

mr->upload_filename is not meant to be used outside the post iterator but needs to be somewhere, where we can hand this to the post_iterator.

So you are correct, after the post iteration is complete the mr->upload_filename will have the name of the last uploaded file.
But once the post iteration is finished you should only access the files (and the according paths and infos) via the files map.

src/webserver.cpp
```diff
@@ -527,9 +567,14 @@ MHD_Result webserver::requests_answer_second_step(MHD_Connection* connection, co
 #ifdef DEBUG
         std::cout << "Writing content: " << std::string(upload_data, *upload_data_size) << std::endl;
 #endif  // DEBUG
-        mr->dhr->grow_content(upload_data, *upload_data_size);
+        if (mr->pp == nullptr || put_processed_data_to_content) {
```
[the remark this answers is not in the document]
ctomahogh: No, it's not redundant.

The post iterator is only created from the libmicrohttpd for content of type `multipart/form-data` and `application/x-www-form-urlencoded`. Otherwise the `MHD_create_post_processor` will not set up a post processor and return `nullptr`.

So even if the libhttpserver is used with `put_processed_data_to_content` to `false`, other content for which no post processor is created (i.e. `application/json`) must be put to the content, because the `get_content` of the libhttpserver is the only way to access this content.

So
- `mr->pp == nullptr` is the indication, that it is neither of the above two content types and therefor has to be put to content
- `put_processed_data_to_content` indicates, that data, which is passed to the post processor should additionally be put into the content (which is the default to preserve the current behavior)

src/httpserver/http_request.hpp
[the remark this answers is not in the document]
ctomahogh: Fixed

src/httpserver/http_request.hpp
```diff
@@ -301,6 +317,17 @@ class http_request {
...
+         file_info_s file_info = { file_size, file_system_file_name };
+         files[upload_file_name] = file_info;
+     }
+
```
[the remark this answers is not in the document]
ctomahogh: Pushed by accident, as I introduced this function on the first shot but actually did not used it.
Is removed (and file_system_file_name is only used as type std::string in the remaining code)

src/webserver.cpp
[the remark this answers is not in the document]
LeSpocky: Makes sense.

src/webserver.cpp
ctomahogh: No, it does not (at least if I can trust my tests).

`filename == nullptr` is the indicator, that the currently targeted key in the post_iterator is not an uploaded file. So these parameters will always be put to the arguments.
the `mr->ws->file_upload_target != FILE_UPLOAD_DISK_ONLY` will trigger, that files (where filename is not a nullptr) should be stored to the arguments too.

I did tests with all of the possible file_upload_target values and all parameters, other content (application/json, ...) was all handled as it should.

Yes, a unit test would be helpful. That's on my agenda.

ctomahogh:
Thanks for the feedback.

I already made some fixed and already commented on some of your comments.

First of all I have to admit, that I am not really a C++-Developer, but more into C (I think, that explains some of the stuff I did).
I will work on all of those const and string-vs-char* stuff and send another commit.

Regarding the asprintf, yes, that is not a good way (as the test build showed too). I will try to get rid of this.

Regarding documentation of the options: Yes I am aware of this and this will follow shortly. I just waited for a general approval of the supplied patch, so that I do not go into the wrong direction. But as you both agreed, that this is a feasible approach, I will do the docu.

Regarding unit-tests: I do not really know, how to setup unit tests on this, as we would need file system operations. How can we mock such things with the used unit test framework?

Regarding mkstemp.
The problem with `tmpnam` is, that you cannot specify, in which directory this file should be located. The function will always use the tmp directory (on linux and windows).
So in order to support the option `post_upload_dir` I would divide this according the OS
For linux I would still use mkstemp
For windows a combination of `_mktemp_s` and `s_open_s` 

One problem is, that I do not have a testing (or compiling) environment for windows.

etr:
Sounds good, and thanks again for all the good work.

> Regarding unit-tests: I do not really know, how to setup unit tests on this, as we would need file system operations. How can we mock such things with the used unit test framework?

The test framework doesn't have mocking mechanism.

For files that you might need in input, there are "real" files that are copied into the test root by the configure script (see: https://github.com/etr/libhttpserver/blob/master/configure.ac#L275). You can use those.

For files that you need to verify in output, you should just be able to write within the testdir or the tmp directory.

> One problem is, that I do not have a testing (or compiling) environment for windows.

The automatic build runs on AppVeyor; while that's not the same as running a local windows instance, that should help.

LeSpocky:
On Thu, Feb 10, 2022 at 09:08:08AM -0800, Sebastiano Merlino wrote:
> For files that you might need in input, there are "real" files that are copied into the test root by the configure script (see: https://github.com/etr/libhttpserver/blob/master/configure.ac#L275). You can use those.

By chance I wrote some tests with libcurl today for a different
project, which actually do file upload with that format.  If you're
interested I can change some example code.

src/webserver.cpp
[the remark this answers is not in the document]
ctomahogh: Implemented, but I went for the solution (which is also part of the provided link) to use a static const char in favor of an inline function.
I hope you agree on that

src/http_utils.cpp
ctomahogh: Done with the rework of this function to not use asprintf any more

src/http_utils.cpp
ctomahogh: I could not figure out how to use only strings inside this function, as mkstemp needs a char * and actually changes the content of the parameter.
So I created the template filename as a string and strdup-ed this into a char * which is then given to mkstemp (or _mktemp_s on windows).
Afterwards the changed char * is put into an output string which is then returned.

So at least the concatenation of the directory, the path separator and the template is done with standard string methods.

I hope this is acceptable.

The empty return is also exchanged with a `throw std::exception` (and do a try - catch in the calling function).
Actually I do not really have an idea, which exception to use, so I went for a quite general one.
If you have any suggestion for a better exception, please give me a hint

src/httpserver/http_utils.hpp
ctomahogh: Is const now (input and output)
And yes, I will use cpplint before pushing in the future.

src/webserver.cpp
ctomahogh: I did a few tests on this (with Postman)

The content (to be retrieved with `get_content`) is always correct.

_Upload the same file with different keys_
- Two separate entries in the args map
- Only one file created with doubled content (and only one file in the files map)

_Upload different files with the same key_
- Only one arg with the doubled content
- Two separate files (and two files in the map)

_Upload same file with the same key (this IMHO makes absolutely no sense)_
- Only one arg and one file with duplicated content

This behaviour is actually logical, as the args map is set up by the key and the files map by the filename
Question is probably: What is the expected behavior?

Better solution for the files map would be to set up the map with an unique identifier and always take key and filename into account.

But how can this be changed for the args? As changing the map will change the API and therefor break the compatibility.

src/webserver.cpp
```diff
@@ -464,6 +465,13 @@ MHD_Result webserver::post_iterator(void *cls, enum MHD_ValueKind kind,
 
     struct details::modded_request* mr = (struct details::modded_request*) cls;
 
+if (key) {
+    std::cout << "got key '" << key << "'" << std::endl;
+}
+if (filename) {
+    std::cout << "got filename '" << filename << "'" << std::endl;
+}
+
```
LeSpocky, line 474: Is this debug code slipped in by accident?

src/http_utils.cpp
ctomahogh: Changed this to
`static const std::string upload_filename_template = "libhttpserver.XXXXXX";`
and removed the declaration from http_utils.hpp

src/http_request.cpp
ctomahogh: The attached `&` and `*` are adjusted. Additionally this function takes `const std::string& upload_file_name` as parameter now.

The return value is now const.
Therefor I introduced new setters for the content of the file_info_s as these where previously set in the calling function directly on the returned reference.

examples/file_upload.cpp
```diff
@@ -61,7 +61,7 @@ class file_upload_resource : public httpserver::http_resource {
             </tr>";
 
         std::map<std::string, file_info_s> files = req.get_files();
-        for (std::map<std::string, file_info_s>::iterator it = files.begin(); it != files.end(); it++) {
+        for (std::map<std::string, file_info_s>::iterator it = files.begin(); it != files.end(); ++it) {
```
LeSpocky, line 64: `auto it` should do in modern C++

ctomahogh:
I already added a very basic example for the file upload and storing to the file system last week.

I don't know, whether there should only be some selected examples as part of the library. So if this should not be part of the merge, I can remove it again.
Otherwise feel free to comment on that example too.

Additionally I reworked the map of files to look like
`std::map<std::string, std::map<std::string, file_info_s>> files;`
where the identifier of the outer map is the key (under which the file was uploaded) and the identifier of the inner map is the name of the uploaded file.
This can cope with both use cases
- upload multiple files with the same key (on input fields with 'multiple')
- upload the same file multiple times using different keys

ctomahogh:
I tried to resolve all of the workflow and code analysis errors, but one thing remains, which I don't really now, how I should solve this.

The Codacy Static Code Analyzer complains, that the file_size (inside the map) is never used.
That's of course true, as this is only put into the map, to be available for the application, which uses the library. The example actually reads and displays it.

I can't figure out how to tell the compiler to ignore this (I tried some `__attribute__((used))` but that did not work).

Do you have any advice how to solve this?
Should I implement a getter function for this, so that there is the possibility to read it?
But the question is, is such a getter function really useful? In my opinion, this is not needed and is unlikely to be used by anybody and would purely exist to satisfy the code analysis.

ctomahogh:
@LeSpocky you offered some example code which might help me writing unit tests.
I would appreciate that.
You can share this via email if you prefer that. Or any other way which suites you.

beneschtech:
Just use (void)variablename in the function. Or omit the variable name from
the declaration. This is a great idea and ill have to try it out!

etr:
> I tried to resolve all of the workflow and code analysis errors, but one thing remains, which I don't really now, how I should solve this.
> 
> The Codacy Static Code Analyzer complains, that the file_size (inside the map) is never used. That's of course true, as this is only put into the map, to be available for the application, which uses the library. The example actually reads and displays it.
> 
> I can't figure out how to tell the compiler to ignore this (I tried some `__attribute__((used))` but that did not work).
> 
> Do you have any advice how to solve this? Should I implement a getter function for this, so that there is the possibility to read it? But the question is, is such a getter function really useful? In my opinion, this is not needed and is unlikely to be used by anybody and would purely exist to satisfy the code analysis.

Given that we use that as a return value in the public API, I'd give it additional dignity as a class/struct.

Move it to its own file (hpp & cpp), add getters, and overall make it immutable (setting it on creation and not having setters). You also don't need to add the "_s" to the name; `file_info` is a fine name given that it lives already within the httpserver namespace.

Incidentally, this will also fix the codacy issue, I believe.

ctomahogh:
I changed the file_info into a class (with own hpp and cpp).
But I cannot really make it immutable.

As we only receive the file in the post processor in chunks (in multiple calls of the post_iterator) the file_size has to be increased on every new chunk. So the function `grow_file_size` will be called multiple times and there is no real way of only setting it in the constructor.
So at least the setter function `grow_file_size` will remain.
So I would also rather leave the `set_file_system_file_name` setter function, as this would make it more difficult to handle the map in the post iterator.

Currently I use the function `get_or_create_file_info` which either returns an already existing file_info or create a new one. Afterwards the fact, that `file_system_file_name` is not yet set (meaning the string is empty), indicates, that this is a new entry.
If on creation of `file_info` the `file_system_file_name` would have to be already known, this would have to be rewritten to first search for an existing entry and create it, if it was not found.
As the currently used `return files[key][upload_file_name];` in the function `get_or_create_file_info` takes care of exactly this, I would rather not rewrite that on my own.

So currently my class definition looks like:
```
class file_info {
 public:
     void set_file_system_file_name(const std::string file_system_file_name);
     void grow_file_size(size_t additional_file_size);
     size_t get_file_size() const;
     const std::string get_file_system_file_name() const;

 private:
     size_t _file_size;
     std::string _file_system_file_name;
};
```

If you agree on that (including the getters and setters) I will push it.

ctomahogh:
So, just to ensure, that I got you right.
We would have an internal class `file_info_internal` which will be used including setters but is not available for calling processes.
And an external class `file_info`, which only has a constructor and getters and is available outside the library.

How would this look like in the `http_request` class?
Would this than have two maps?
`std::map<std::string, std::map<std::string, http::file_info_internal>> files_internal;`
and
`std::map<std::string, std::map<std::string, http::file_info>> files;`
And after the post_iterator is done and the `files_internal` map is set up, copy this to the `files` map, which can then be retrieved via the `get_files` function.

The point is, that we can only be sure, that the data inside the `files_internal` class is completely there, after the post_iterator is finished.
So this copy to the external map would happen in `webserver::complete_request`.

Did I get you right?

etr:
Yeah, basically you can store the temporary map in the `modded_request` object that gets passed to the post_iterator at each iteration. Once you are done, you can move the content to the new map within the http_request.

To be fair, given the complexity and the fact that we already do the same in `http_request`, it might be worth just keeping those setter methods private in the `file_info` class and have the `webserver` class marked as friend. While I am not in love with this either, I think it is the better of the two at this point and should lead to more readable code.

Sorry for the rapid change of direction, but I hadn't initially realized a tricky it was given the multiple loops in the post_iterator.

Hope it makes sense.

src/http_utils.cpp
```diff
@@ -193,6 +198,14 @@ const char* http_utils::http_post_encoding_multipart_formdata = MHD_HTTP_POST_EN
 const char* http_utils::application_octet_stream = "application/octet-stream";
 const char* http_utils::text_plain = "text/plain";
 
+static const char* upload_filename_template = "libhttpserver.XXXXXX";
+
+#ifdef _WIN32
```
etr, line 203: nit: mostly for consistency, can we use `if defined(` instead of `ifdef` ?

src/http_utils.cpp
```diff
@@ -193,6 +198,14 @@ const char* http_utils::http_post_encoding_multipart_formdata = MHD_HTTP_POST_EN
 const char* http_utils::application_octet_stream = "application/octet-stream";
 const char* http_utils::text_plain = "text/plain";
 
+static const char* upload_filename_template = "libhttpserver.XXXXXX";
```
etr, line 201: let's capitalize the field name given it is static const

src/http_utils.cpp
```diff
@@ -193,6 +198,14 @@ const char* http_utils::http_post_encoding_multipart_formdata = MHD_HTTP_POST_EN
...
+
+#ifdef _WIN32
+    static const char path_separator = '\\';
```
etr, line 204: same as above, let's use a name in all caps.

src/http_utils.cpp
```diff
@@ -216,6 +229,39 @@ std::string http_utils::standardize_url(const std::string& url) {
...
+    std::string filename = directory + path_separator + upload_filename_template;
+    char *template_filename = strdup(filename.c_str());
+    int fd = 0;
+
+#if defined(_WIN32)
+    if (0 != _mktemp_s(template_filename, filename.size() + 1)) {
+        free(template_filename);
+        throw std::exception();
+    }
+    if (0 != _sopen_s(&fd, template_filename, _O_CREAT | _O_EXCL | _O_NOINHERIT, _SH_DENYNO, _S_IREAD | _S_IWRITE)) {
```
etr, line 242: if we are just trying to read the file to verify it is created correctly, we can just use `ifstream`

src/http_utils.cpp
etr: No issues with the string.

For the exception, let's create a custom exception so that we can then catch that (and we don't incur in the risk of catching exceptions raised by other parts of the code)

src/http_utils.cpp
```diff
@@ -216,6 +229,39 @@ std::string http_utils::standardize_url(const std::string& url) {
...
+    _close(fd);
+#else
+    fd = mkstemp(template_filename);
+
+    if (fd == -1) {
+        free(template_filename);
+        throw std::exception();
+    }
+    close(fd);
+#endif
```
etr, line 259: add comment after the #endif and the else with the value of the ifdef // _WIN32

src/httpserver/file_info.hpp
```diff
@@ -0,0 +1,55 @@
...
+#include <string>
+
+namespace httpserver {
+class webserver;
+
+namespace http {
+
+class file_info {
+ public:
+     size_t get_file_size() const;
```
etr, line 37: nit: can we make the output `const` ?

src/webserver.cpp
```diff
@@ -454,14 +458,57 @@ MHD_Result webserver::post_iterator(void *cls, enum MHD_ValueKind kind,
...
-    return MHD_YES;
+
+    try {
+        if (filename == nullptr || mr->ws->file_upload_target != FILE_UPLOAD_DISK_ONLY) {
+            mr->dhr->set_arg(key, mr->dhr->get_arg(key) + std::string(data, size));
+        }
+
+        if (filename && *filename != '\0' && mr->ws->file_upload_target != FILE_UPLOAD_MEMORY_ONLY) {
+            // either get the existing file info struct or create a new one in the file map
+            http::file_info &file = mr->dhr->get_or_create_file_info(key, filename);
```
etr, line 474: nit: move `&` close to `file_info`

src/webserver.cpp
```diff
@@ -454,14 +458,57 @@ MHD_Result webserver::post_iterator(void *cls, enum MHD_ValueKind kind,
...
+                    file.set_file_system_file_name(http_utils::generate_random_upload_filename(mr->ws->post_upload_dir));
+                } else {
+                    file.set_file_system_file_name(mr->ws->post_upload_dir + "/" + std::string(filename));
+                }
+                // to not append to an already existing file, delete an already existing file
+                unlink(file.get_file_system_file_name().c_str());
+            }
+
+            // if multiple files are uploaded, a different filename indicates the start of a new file, so close the previous one
+            if (mr->upload_filename.empty() || mr->upload_key.empty() || 0 != strcmp(filename, mr->upload_filename.c_str()) || 0 != strcmp(key, mr->upload_key.c_str())) {
```
etr, line 488: nit: let's break this line up on multiple lines

src/webserver.cpp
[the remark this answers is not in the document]
etr: Given the question from @LeSpocky, let's add a comment clarifying what the if statement is doing.

etr (commented):
Overall, this is shaping up really well. I guess at this point we only miss unit tests and we are done.

I added a few comments to the PR itself.

Again, thanks for all the work you are doing here.

src/http_utils.cpp
ctomahogh: That's the point where the file is actually created in the file system.
`_mktemp_s()` just generates a file name but does not create the file in the file system.

If this call fails, the file could not be created.

Of course, there are other commands to create a file, but I went for `_sopen_s()` as with this function you can specify the flags (like _O_CREAT, ...) in detail.

src/httpserver/http_request.hpp
```diff
@@ -135,6 +136,22 @@ class http_request {
      **/
      const std::map<std::string, std::string, http::arg_comparator> get_args() const;
 
+     /**
+      * Method to get or create a file info struct in the map if the provided filename is already in the map
+      * return the exiting file info struct, otherwise create one in the map and return it.
+      * @param upload_file_name the file name the user uploaded (this is the identifier for the map entry)
+      * @result a file info struct file_info_s
+     **/
+     http::file_info& get_or_create_file_info(const std::string &key, const std::string& upload_file_name);
```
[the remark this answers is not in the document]
ctomahogh: As far as I see it, the output cannot be const, as in the using function (`post_iterator()`) the setters of the file_info class are called on the returned file_info.

src/http_utils.cpp
```diff
@@ -230,33 +230,33 @@ std::string http_utils::standardize_url(const std::string& url) {
...
 
     if (fd == -1) {
         free(template_filename);
-        throw std::exception();
+        throw generateFilenameException();
     }
     close(fd);
-#endif
+#endif  // _WIN32
     std::string ret_filename = std::string(template_filename);
```
etr, line 260: If `template_filename` is a `char*` you might not need to wrap it into an `std::string`; the `operator=` does the job already for you.

src/webserver.cpp
```diff
@@ -465,27 +464,34 @@ MHD_Result webserver::post_iterator(void *cls, enum MHD_ValueKind kind,
...
-        if (filename == nullptr || mr->ws->file_upload_target != FILE_UPLOAD_DISK_ONLY) {
+        if (filename == nullptr ||
+            mr->ws->file_upload_target != FILE_UPLOAD_DISK_ONLY) {
             mr->dhr->set_arg(key, mr->dhr->get_arg(key) + std::string(data, size));
         }
 
-        if (filename && *filename != '\0' && mr->ws->file_upload_target != FILE_UPLOAD_MEMORY_ONLY) {
+        if (filename &&
+            *filename != '\0' &&
+            mr->ws->file_upload_target != FILE_UPLOAD_MEMORY_ONLY) {
```
etr, line 474: nit: The length of this line isn't long enough to justify the break. You can keep it on one line for easier readability

src/http_utils.cpp
etr: If there is something highly specific we are trying to do here with the explicit params, let's keep `_sopen_s` but explain what we are trying to achieve in a comment so that it is immediately clear.
If there isn't, in general I'd prefer using C++ constructs (ofstream/ifstream) over C ones; they tend to port more easily across platforms.

etr:
The latest failure I see on cppcheck is due to an update of cppcheck version on `ubuntu-latest` used by GitHub actions. If you pull from master, that should solve itself.

Let me know if there is any help/support you need from me on the unit tests.

test/unit/http_utils_test.cpp
```diff
@@ -31,7 +31,7 @@
 #endif
 
 #include <cstdio>
-
+#include <filesystem>
```
LeSpocky, line 34: This requires C++17, not sure what @etr thinks about requiring C++17 in libhttpserver.  Current requirement is C++14 for the whole project.  Not sure if there should be different requirements for lib and tests?

test/unit/http_utils_test.cpp
```diff
@@ -154,6 +154,29 @@ LT_BEGIN_AUTO_TEST(http_utils_suite, standardize_url)
...
 
+LT_BEGIN_AUTO_TEST(http_utils_suite, generate_random_upload_filename)
+    const char* UPLOAD_FILENAME_TEMPLATE = "libhttpserver.XXXXXX";
+
+    #if defined(_WIN32)
+        const char PATH_SEPARATOR = '\\';
+    #else  // _WIN32
+        const char PATH_SEPARATOR = '/';
+    #endif  // _WIN32
+
```
LeSpocky, line 165: Okay, if <filesystem> is used from C++17, then this is not needed anymore.  There's `std::filesystem::path` then which has operators for building paths, so users don't have to worry about / or \ anymore.  See https://en.cppreference.com/w/cpp/filesystem/path for reference.

ctomahogh:
Some comments/questions on the unit test for `generate_random_upload_filename`

- I currently duplicated the `PATH_SEPARATOR` and the `UPLOAD_FILENAME_TEMPLATE`
I do not like duplicating it, but the only idea I have is, to make the strings public again. In the library (if we ignore the unit tests) both strings are only used inside `http_utils.cpp` and are therefor intentionally not visible outside of this .cpp file. What would be the desired way here
- the tested function not only returns an unique filename, but also creates this file. This is currently only done in the current directory. Additionally this file gets deleted afterwards, to not fill the file system due to the test. 
If wished, I could add a check, whether the unlink actually succeeded. But as this is not really a test for the function under test, I did not do this up to now
- Testing with other directories might be possible within this unit test, but this would require to create directories (and remove them afterwards)

Further question regarding tests
- I cannot write unit tests for `file_info.cpp` (or `file_info.hpp`) as the only class, able to set values in this class, is the class webserver (as it is a friend of this class).
- I don't think the `post_iterator` and the new files map can easily be tested with unit tests
- So I would rather write a new integration test, which tests
-- the upload of files
-- checks the created files map
-- checks the existence of the files
-- checks the existence of the file in the content
-- checks the args map.
This would then test the `post_iterator`, the files map and the getters of `file_info`

test/unit/http_utils_test.cpp
ctomahogh: Oh, I missed that.
I just searched for the easiest portable way to check a file existence.

On linux this check would actually not be necessary, as mkstemp always creates a file and would fail, if it could not.
But on windows the creation of the file name and creation of the file are two separate steps in the function, so in my opinion this check should stay.

I will search for a easy C++14 way of doing that.

etr: thanks.

In general, I am not against a move to strict C++17, but this would be a bit small to justify that. Filesystem is doubly unfortunate because it would force C++ to version >8 and Clang >7.

In practice, you can check for file existence the same way it is done here: https://github.com/etr/libhttpserver/blob/master/src/file_response.cpp

LeSpocky:
>     * I currently duplicated the `PATH_SEPARATOR` and the `UPLOAD_FILENAME_TEMPLATE`
>       I do not like duplicating it, but the only idea I have is, to make the strings public again. In the library (if we ignore the unit tests) both strings are only used inside `http_utils.cpp` and are therefor intentionally not visible outside of this .cpp file. What would be the desired way here

In other projects I usually put such stuff into a private header file, which I can include in the .c or .cpp files, but which is not installed and distributed to the user of the lib.

etr:
> So I would rather write a new integration test, which tests
-- the upload of files
-- checks the created files map
-- checks the existence of the files
-- checks the existence of the file in the content
-- checks the args map.
This would then test the post_iterator, the files map and the getters of file_info

That makes sense to me, and it is in line with what we've been doing with the library in general when dealing with high-level functionalities.

ctomahogh:
> > * I currently duplicated the `PATH_SEPARATOR` and the `UPLOAD_FILENAME_TEMPLATE`
> >   I do not like duplicating it, but the only idea I have is, to make the strings public again. In the library (if we ignore the unit tests) both strings are only used inside `http_utils.cpp` and are therefor intentionally not visible outside of this .cpp file. What would be the desired way here
> 
> In other projects I usually put such stuff into a private header file, which I can include in the .c or .cpp files, but which is not installed and distributed to the user of the lib.

Really only in a header file?
Shouldn't this be declared in a .hpp file and defined in a .cpp file like

file_name_constants.hpp:
```
extern const char* UPLOAD_FILENAME_TEMPLATE;
```
file_name_constants.cpp:
```
const char* UPLOAD_FILENAME_TEMPLATE = "libhttpserver.XXXXXX";
```

At least that's (in my opinion) the best way in C. But honestly I don't know, whether other ways of accomplishing this are acceptable and common practice in C++.

Either way, should we really do this for only two constant strings?
Or are there already some other constant string in libhttpserver, which are not exported? I could not find any, but maybe I missed something?
Most constant strings are part of http_utils and therefor visible externally.
But would this really be a problem? As the filename template (which defines the basic filename) is quite important for a user.

To sum up three options:
1.) Make it part of http_utils again (as it was at the very beginning)
2.) Make only one header file with declaration and definition (could you please provide an example, as I do not know the common C++ way for this
3.) Make one .hpp and one .cpp file for this (could you then state, in which subfolders such "internal" files should reside?)

I will follow your wish.

LeSpocky:
> > > * I currently duplicated the `PATH_SEPARATOR` and the `UPLOAD_FILENAME_TEMPLATE`
> > >   I do not like duplicating it, but the only idea I have is, to make the strings public again. In the library (if we ignore the unit tests) both strings are only used inside `http_utils.cpp` and are therefor intentionally not visible outside of this .cpp file. What would be the desired way here
> > 
> > 
> > In other projects I usually put such stuff into a private header file, which I can include in the .c or .cpp files, but which is not installed and distributed to the user of the lib.
> 
> Really only in a header file? Shouldn't this be declared in a .hpp file and defined in a .cpp file like
> 
> file_name_constants.hpp:
> 
> ```
> extern const char* UPLOAD_FILENAME_TEMPLATE;
> ```
> 
> file_name_constants.cpp:
> 
> ```
> const char* UPLOAD_FILENAME_TEMPLATE = "libhttpserver.XXXXXX";
> ```
> 
> At least that's (in my opinion) the best way in C. But honestly I don't know, whether other ways of accomplishing this are acceptable and common practice in C++.

You're right, I did not look closely. From the top of my head I remember equal string literals are not put twice into binary but deduplicated. I guess that's done by the linker, but not sure and also not sure if it's global over all object files or per object. And given that I was irritated about assigning that to a pointer in the first place. Why not just do 

```
#define UPLOAD_FILENAME_TEMPLATE "libhttpserver.XXXXXX"
```

or is that not possible because it's used as template for `mkstemp()` (?) later? That would be usable from only a header file.

src/webserver.cpp
```diff
@@ -454,14 +457,61 @@ MHD_Result webserver::post_iterator(void *cls, enum MHD_ValueKind kind,
         const char *transfer_encoding, const char *data, uint64_t off, size_t size) {
     // Parameter needed to respect MHD interface, but not needed here.
     std::ignore = kind;
-    std::ignore = filename;
     std::ignore = content_type;
     std::ignore = transfer_encoding;
```
etr, line 461: While re-reading this change I was wonder whether we should include these two parameters as part of the `file_info` structure as well to make it possible for the clients to know the content type and encoding of the file. What do you think?

It would likely also align with the issue that originated the whole initial discussion: https://github.com/etr/libhttpserver/issues/229

etr:
> > > > * I currently duplicated the `PATH_SEPARATOR` and the `UPLOAD_FILENAME_TEMPLATE`
> > > >   I do not like duplicating it, but the only idea I have is, to make the strings public again. In the library (if we ignore the unit tests) both strings are only used inside `http_utils.cpp` and are therefor intentionally not visible outside of this .cpp file. What would be the desired way here
> > > 
> > > 
> > > In other projects I usually put such stuff into a private header file, which I can include in the .c or .cpp files, but which is not installed and distributed to the user of the lib.
> > 
> > 
> > Really only in a header file? Shouldn't this be declared in a .hpp file and defined in a .cpp file like
> > file_name_constants.hpp:
> > ```
> > extern const char* UPLOAD_FILENAME_TEMPLATE;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > file_name_constants.cpp:
> > ```
> > const char* UPLOAD_FILENAME_TEMPLATE = "libhttpserver.XXXXXX";
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > At least that's (in my opinion) the best way in C. But honestly I don't know, whether other ways of accomplishing this are acceptable and common practice in C++.
> 
> You're right, I did not look closely. From the top of my head I remember equal string literals are not put twice into binary but deduplicated. I guess that's done by the linker, but not sure and also not sure if it's global over all object files or per object. And given that I was irritated about assigning that to a pointer in the first place. Why not just do
> 
> ```
> #define UPLOAD_FILENAME_TEMPLATE "libhttpserver.XXXXXX"
> ```
> 
> or is that not possible because it's used as template for `mkstemp()` (?) later? That would be usable from only a header file.

To be fair, I am not too fussed about the repetition in this case (as it goes through prod and tests and the worst case result in this case are failing tests if we change the name, but not falsely succeeding tests). Nonetheless, I understand the sentiment. I wouldn't over-complicate it, though. I'd be fine with them living in http_utils and being visible - they are const, anyway.

If we want to give specific indication that these are indeed private to the library and shouldn't be used from the outside, the convention has been to put them within a separate namespace (`httpserver::details`). We already assume the clients won't use anything from within this namespace.

src/webserver.cpp
ctomahogh: Done, added content_type and transfer_encoding.

One thing to note:
These two are only available "if known" (as written in libmicrohttpd documentation).
In my tests the transfer_encoding was always nullptr and therefor not set.
The content_type I tested successfully (and integrated in the file_upload example)

LeSpocky:
> @LeSpocky you offered some example code which might help me writing unit tests. I would appreciate that. You can share this via email if you prefer that. Or any other way which suites you.

This is from our applications unit tests, redacted so you won't be able to compile it, and it's using a different unit test framework, but you get an idea how to create those multipart form data requests with libcurl: 

https://gist.github.com/LeSpocky/b35b33b835116bbdb16fad7ba3c0c34e

It's basically using those `curl_mime_XXX()` functions, quite easy to use actually.

ctomahogh:
I just added an integration test for the file upload

Please review, as I don't really know, how I should (or could) improve this.
As this integration test tests different configuration options of the webserver, there is some code (especially the tests) duplicated.
Please advice, if I could improve the code here.

test/integ/file_upload.cpp
```diff
@@ -0,0 +1,263 @@
...
+                                         stringstream* content) {
+         this->args = args;
+         this->content = content;
+         this->files = files;
+     }
+
+     const shared_ptr<http_response> render_POST(const http_request& req) {
+         (*content) << req.get_content();
+         (*args) = req.get_args();
+         (*files) = req.get_files();
```
etr, line 85: Can we manage these class attributes as internal values and not use pointers (better with setters/getters, acceptable with a public interface)?

Ideally, from the outside, you can use the getters.

test/integ/file_upload.cpp
```diff
@@ -0,0 +1,263 @@
...
+    std::map<std::string, std::map<std::string, httpserver::http::file_info>>::iterator file_key = files.begin();
+    LT_CHECK_EQ(file_key->first, TEST_KEY);
+    LT_CHECK_EQ(file_key->second.size(), 1);
+    std::map<std::string, httpserver::http::file_info>::iterator file = file_key->second.begin();
+    LT_CHECK_EQ(file->first, TEST_CONTENT_FILENAME);
+    LT_CHECK_EQ(file->second.get_file_size(), TEST_CONTENT_SIZE);
+    LT_CHECK_EQ(file->second.get_content_type(), httpserver::http::http_utils::application_octet_stream);
+
+    string expected_filename = upload_directory + httpserver::http::http_utils::path_separator + httpserver::http::http_utils::upload_filename_template;
+    LT_CHECK_EQ(file->second.get_file_system_file_name().substr(0, file->second.get_file_system_file_name().size() - 6), expected_filename.substr(0, expected_filename.size() - 6));
```
etr, line 144: nit: these lines are a bit long; let's wrap them

test/integ/file_upload.cpp
```diff
@@ -0,0 +1,263 @@
...
```
etr, line 187: nit: these lines are a bit long; let's wrap them

etr (commented):
Thanks for adding the tests.

There are a few tests that I'd additionally do on this.

(1) A test where we provide both a file to upload and other parameters to parse.
(2) A test where we provide multiple files in input.
(3) A test where we upload through PUT (instead of POST)
(4) A test where we prove the post-processor works on the uploaded file.

ctomahogh:
Is the current error for the AppVeyor my fault?

I did not knowingly change anything to the YAML file. The file is (according to the repo) still there and untouched since 7 days (which was a merge of the current master).

Or is this only a temporary sporadic issue and the workflow just needs to be restarted (which I cannot do)?

etr:
> Is the current error for the AppVeyor my fault?
> 
> I did not knowingly change anything to the YAML file. The file is (according to the repo) still there and untouched since 7 days (which was a merge of the current master).
> 
> Or is this only a temporary sporadic issue and the workflow just needs to be restarted (which I cannot do)?

Not your fault at all, I think AppVeyor itself was having issues with Github permissions yesterday

test/integ/file_upload.cpp
```diff
@@ -0,0 +1,491 @@
...
+class print_file_upload_resource : public http_resource {
+ public:
+     const shared_ptr<http_response> render_POST(const http_request& req) {
+         content = req.get_content();
+         args = req.get_args();
+         files = req.get_files();
+         shared_ptr<string_response> hresp(new string_response("OK", 201, "text/plain"));
+         return hresp;
+     }
+     const shared_ptr<http_response> render_PUT(const http_request& req) {
```
etr, line 141: nit: empty line before this

test/integ/file_upload.cpp
```diff
@@ -0,0 +1,491 @@
...
+    ws->register_resource("upload", &resource);
+
+    CURLcode res = send_file_to_webserver(false, false);
+    LT_ASSERT_EQ(res, 0);
+
+    string actual_content = resource.get_content();
+    LT_CHECK_EQ(actual_content.find(FILENAME_IN_GET_CONTENT) != string::npos, true);
+    LT_CHECK_EQ(actual_content.find(TEST_CONTENT) != string::npos, true);
+
+    auto args = resource.get_args();
```
etr, line 197: nit: let's avoid `auto` in this context - it makes it harder to understand the type used (same for all the others in this code).

etr:
Only a couple of nitpicky comments, but I think we are done for the rest here.

etr:
Approved the PR.

Thanks for all the great work and the amount of effort you put into getting this done. This is a great piece of improvement to the library.

ctomahogh:
Great news.

I am glad, that I could contribute.
Thanks for all the review and help you provided for a (mainly) C-Programmer ;-)

Issue #258, opened 2022-02-10
[BUG] libmicrohttpd v0.9.75 compatibility


LeSpocky:
### Prerequisites

* [x] Put an X between the brackets on this line if you have checked that your issue isn't already filed: https://github.com/search?l=&q=repo%3Aetr%2Flibhttpserver&type=Issues

### Description

libmicrohttpd v0.9.74 marks some definitions as deprecated, which leads to compiler warnings when using them, which libhttpserver does.  This leads to build failure when calling configure with `--enable-debug` because that adds `-Werror` to compiler flags.

### Steps to Reproduce

1. Build libmicrohttpd v0.9.74 or v0.9.75
2. Call `../configure` with `--enable-debug`
3. Call `make`

**Expected behavior:** Successful build.

**Actual behavior:** Build fails.

**Reproduces how often:** Always.

### Versions

* OS version: Debian GNU/Linux 10 (buster)
* libhttpserver version: master
* libmicrohttpd version: v0.9.75, compiled

If you have problems during build:
* Compiler version: gcc (Debian 8.3.0-6) 8.3.0
* autotools version: ??

### Additional Information

The deprecated definitions are these:

- MHD_HTTP_UNORDERED_COLLECTION
- MHD_HTTP_PAYLOAD_TOO_LARGE
- MHD_HTTP_UNPROCESSABLE_ENTITY

Full build output:

```
% make
make  all-recursive
make[1]: Verzeichnis „/home/adahl/src/libhttpserver/build“ wird betreten
Making all in src
make[2]: Verzeichnis „/home/adahl/src/libhttpserver/build/src“ wird betreten
/bin/bash ../libtool  --tag=CXX   --mode=compile g++ -DHAVE_CONFIG_H -I. -I../../src -I..  -I../ -I../../src/httpserver/ -I/home/adahl/build/sysroot/http/include -DDEBUG -g -Wall -Wextra -Werror -pedantic -std=c++14 -Wno-unused-command-line-argument -O0 -fPIC -Wall  -DUSE_FASTOPEN -DHTTPSERVER_COMPILATION -D_REENTRANT   -MT libhttpserver_la-http_utils.lo -MD -MP -MF .deps/libhttpserver_la-http_utils.Tpo -c -o libhttpserver_la-http_utils.lo `test -f 'http_utils.cpp' || echo '../../src/'`http_utils.cpp
[...]

Issue #176, opened 2020-02-06
[BUG] Use of deprecated MHD_HTTP_UNORDERED_COLLECTION 


bcsgh:
### Prerequisites

* [ X] Put an X between the brackets on this line if you have checked that your issue isn't already filed: https://github.com/search?l=&q=repo%3Aetr%2Flibhttpserver&type=Issues

### Description

Used here:
https://github.com/etr/libhttpserver/blob/c91e2417cce02c259250e81494539f7848ac117a/src/http_utils.cpp#L103

however libmicrohttpd defines it like:

/** @deprecated */
#define MHD_HTTP_UNORDERED_COLLECTION \
  _MHD_DEPR_IN_MACRO ( \
    "Value MHD_HTTP_UNORDERED_COLLECTION is deprecated as it was removed from RFC") \
  425
https://git.gnunet.org/libmicrohttpd.git/tree/src/include/microhttpd.h?h=v0.9.69&id=3a661f8675e4312f5c02339a0dcd44bf0a5fcef5#n498

Note 1: This seems to have been chnaged c2019/07
https://lists.gnu.org/archive/html/gnunet-svn/2019-06/msg00286.html

Note 2: I was able to work around this by building that file with -DMHD_NO_DEPRECATION

### Steps to Reproduce

1. compile with -Werror

**Expected behavior:** [What you expect to happen]

Build without warnings or errors

**Actual behavior:** [What actually happens]

Warning/Error -> "Value MHD_HTTP_UNORDERED_COLLECTION is deprecated as it was removed from RFC"

**Reproduces how often:** 100%

### Versions

See link above for srv versions, OS shouldn't matter. 

If you have problems during build:
* Compiler version: GCC

etr:
Thanks for notifying. It might be about time to update those constants in bulk. I will post a fix in the next few days.

LeSpocky:
FWIW: this definition was deprecated with libmicrohttpd v0.9.63-113-gbadb814e and thus was released with libmicrohttpd v0.9.64 in June 2019. I did not notice this because I was mostly building against libmicrohttpd v0.9.62 from Debian 10 (buster), which is currently the _oldstable_ Debian release.

bcsgh:
Thank!

PR #259, opened 2022-02-10
Fix warnings caused by libmicrohttpd v0.9.74

LeSpocky:
### Identify the Bug

#176, #258 

### Description of the Change

This adds the same definitions as introduced with libmicrohttpd v0.9.74 if built against earlier versions of libmicrohttpd, at least for two out of three.  The other was not renamed in libmicrohttpd, but "removed".

### Alternate Designs

Do not rely on libmicrohttpd, but use integer literals directly.

### Possible Drawbacks

More compat code.

### Verification Process

Build against libmicrohttpd v0.9.75, v0.9.74, and v0.9.73.

### Release Notes

- Remove `http_utils::http_unordered_collection`

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -56,9 +56,9 @@ jobs:
 
     - name: Install libmicrohttpd dependency
       run: |
-        curl https://s3.amazonaws.com/libhttpserver/libmicrohttpd_releases/libmicrohttpd-0.9.59.tar.gz -o libmicrohttpd-0.9.59.tar.gz ;
-        tar -xzf libmicrohttpd-0.9.59.tar.gz ;
-        cd libmicrohttpd-0.9.59 ;
+        curl https://s3.amazonaws.com/libhttpserver/libmicrohttpd_releases/libmicrohttpd-0.9.64.tar.gz -o libmicrohttpd-0.9.64.tar.gz ;
+        tar -xzf libmicrohttpd-0.9.64.tar.gz ;
+        cd libmicrohttpd-0.9.64 ;
         ./configure --disable-examples ;
         make ;
         sudo make install ;
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -36,6 +36,15 @@ jobs:
     - run: git checkout HEAD^2
       if: ${{ github.event_name == 'pull_request' }}
 
+    - name: Install libmicrohttpd dependency
+      run: |
+        curl https://s3.amazonaws.com/libhttpserver/libmicrohttpd_releases/libmicrohttpd-0.9.64.tar.gz -o libmicrohttpd-0.9.64.tar.gz ;
+        tar -xzf libmicrohttpd-0.9.64.tar.gz ;
+        cd libmicrohttpd-0.9.64 ;
+        ./configure --disable-examples ;
+        make ;
+        sudo make install ;
+
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
       uses: github/codeql-action/init@v1
@@ -54,20 +63,11 @@ jobs:
     #    and modify them (or add more) to build your code if your project
     #    uses a compiled language
 
-    - name: Install libmicrohttpd dependency
-      run: |
-        curl https://s3.amazonaws.com/libhttpserver/libmicrohttpd_releases/libmicrohttpd-0.9.64.tar.gz -o libmicrohttpd-0.9.64.tar.gz ;
-        tar -xzf libmicrohttpd-0.9.64.tar.gz ;
-        cd libmicrohttpd-0.9.64 ;
-        ./configure --disable-examples ;
-        make ;
-        sudo make install ;
-
     - name: Manual steps to build the library
       run: |
         ./bootstrap ;
         ./configure --enable-same-directory-build;
         make ;
- 
+
     - name: Perform CodeQL Analysis
       uses: github/codeql-action/analyze@v1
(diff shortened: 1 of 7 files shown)


LeSpocky:
Not sure how to handle `MHD_HTTP_UNORDERED_COLLECTION`, because there's no replacement for that in libmicrohttpd.  Reason for deprecation is that status was removed from updated RFC.

etr:
> Marked this as draft, because I'm not sure how to handle `MHD_HTTP_UNORDERED_COLLECTION`. There's no replacement for that in libmicrohttpd like for the other two definitions. Reason for deprecation is that status was removed from updated RFC.

I'd be tempted to just remove it. There was already a discussion abouit it in a different thread

etr:
Just to add to the above, our current minimum bar for libmicrohttpd is 0.9.53; so you can just remove anything that wasn't there in that version (no need to keep it in optional compilation to support any version of libmicrohttpd < 0.9.53).

LeSpocky:
> > Marked this as draft, because I'm not sure how to handle `MHD_HTTP_UNORDERED_COLLECTION`. There's no replacement for that in libmicrohttpd like for the other two definitions. Reason for deprecation is that status was removed from updated RFC.
> 
> I'd be tempted to just remove it. There was already a discussion abouit it in a different thread

That was #176 as far as I can see?

etr:
Yep - you are correct

etr:
Let's just fix the README in the requirements to point to libmicrohttpd `0.9.63` instead of `0.9.53`

LeSpocky:
> Let's just fix the README in the requirements to point to libmicrohttpd `0.9.63` instead of `0.9.53`

According to my analysis in #176 (https://github.com/etr/libhttpserver/issues/176#issuecomment-1046506415) it would be 0.9.64 … should this just go to README.md or be a hard requirement against libmicrohttpd at build time?

etr:
You are right, and it should be both - the configure.ac has a check for the minimum version of libmicrohttpd; we'll just need to update that.

etr:
I've added the 0.9.64 version to the private S3. You will have to update both appveyor and github actions to point to that version.

LeSpocky:
The CodeQL test does not fail on libhttpserver but on libmicrohttpd:

> Check failure on line 1942 in libmicrohttpd-0.9.64/src/microhttpd/connection.c

etr:
That's annoying. We should skip the libmicrohttpd directory from CodeQL (as we don't have control of that).

It will require us introducing a custom codeql config file as in: https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#using-a-custom-configuration-file; and specify exclude paths as in: https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#specifying-directories-to-scan

LeSpocky:
> Analyze (cpp)
> The "paths"/"paths-ignore" fields of the config only have effect for JavaScript, Python, and Ruby

🙄

etr:
> > Analyze (cpp)
> > The "paths"/"paths-ignore" fields of the config only have effect for JavaScript, Python, and Ruby
> 
> 🙄
> 
> Reading again in the docs:
> 
> > For compiled languages, if you want to limit code scanning to specific directories in your project, you must specify appropriate build steps in the workflow. The commands you need to use to exclude a directory from the build will depend on your build system. For more information, see "[Configuring the CodeQL workflow for compiled languages](https://docs.github.com/en/code-security/secure-coding/configuring-the-codeql-workflow-for-compiled-languages#adding-build-steps-for-a-compiled-language)."
> 
> However that gives me no hint and https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/troubleshooting-the-codeql-workflow is of no help either.
> 
> This reads for me like: “only build what you want to have analyzed” and that would make it impossible because libmicrohttpd is needed. Maybe one can install it from the Ubuntu repositories instead of building it? https://packages.ubuntu.com/impish/libmicrohttpd-dev

Yeah, the only two ways that come to mind are: (1) remove libmicrohttpd build directory before running CodeQL (not sure it would work); (2) using indeed a system-installed version.

LeSpocky:
> > This reads for me like: “only build what you want to have analyzed” and that would make it impossible because libmicrohttpd is needed. Maybe one can install it from the Ubuntu repositories instead of building it? https://packages.ubuntu.com/impish/libmicrohttpd-dev
> 
> Yeah, the only two ways that come to mind are: (1) remove libmicrohttpd build directory before running CodeQL (not sure it would work); (2) using indeed a system-installed version.

1. Maybe this is possible by setting up interdependent jobs? Or just move the step installing the dependency upwards before the CodeQL init step? Will try that.
2. For installing things I found this: https://docs.github.com/en/actions/using-github-hosted-runners/customizing-github-hosted-runners

etr:
The inversion seems to have worked.

etr:
I guess at this point is just a matter of aligning the configure.ac and the README

LeSpocky:
> I guess at this point is just a matter of aligning the configure.ac and the README

So changes to CodeQL workflow should go to another PR before this then?

etr:
> > I guess at this point is just a matter of aligning the configure.ac and the README
> 
> So changes to CodeQL workflow should go to another PR before this then?
> 
> And it's not only about removing MHD_HTTP_UNORDERED_COLLECTION and bumping the required version. What about the two other definitions?

I might be missing what you mean. You already have everything here (removal of MHD_HTTP_UNORDRED_COLLECTION and CodeQL changes); all I was saying was to do also the changes to the README and the configure.ac. Does it make sense?

LeSpocky:
You are right. I somehow thought I had already done this. Had a look at the changes again and it's still missing. Oops. Will add it later today.

Issue #264, opened 2022-03-23
Add a hook or callback for cleaning up on disk file uploads


LeSpocky:
**Is your feature request related to a problem? Please describe.**

Since 3fd99c87b8ea999df593060999be59786e844bba it's possible to let libhttpserver store files uploaded with forms (mime-type *multipart/form-data*) to disk, see the pull request #257 for reference.  This whole new feature request is about cases where file upload target is set to `FILE_UPLOAD_DISK_ONLY` or `FILE_UPLOAD_MEMORY_AND_DISK`. 

In certain circumstances it is necessary for an application to not store those files permanently, but remove them from disk again, after processing.  Doing that in a render method associated with a registered resource is possible.  However the file is always stored to disk, regardless if that render method is called at all.  Multiple code paths in `webserver::finalize_answer()`, mostly error cases, are possible leading to `mr->dhrs` assigned by something else than the render callback supplied by the application.  Simplest case is sending a (possibly forged) request to a not registered resource (HTTP status 404), but others are possible, too.  That would mean the cleanup in your render method is not executed.

The actual target we use libhttpserver on is an embedded device with usable RAM smaller than files uploaded, so we have to store those files on disk.  Those files can and should be removed again on every request, otherwise the filesystem would fill up quickly.  This is especially true for manipulated requests with bad intent.  A full filesystem can be considered as a denial of service situation.

**Describe why the feature or enhancement you are proposing fits the library.**

I propose some extension to libhttpserver for cleaning up all uploaded files of a request at the end of request processing.  Thinking of the possibility of optionally hooking up a callback function provided by the application, where user can decide how to cleanup, and call that in an appropriate place.

**Describe the solution you'd like**

Extend the `create_webserver` options by a function to set a custom cleanup function.  That function would not be called if not set by the application.  However if set it could be executed next to `MHD_destroy_response(raw_response);` at the end of `webserver::finalize_answer()`? 

**Describe alternatives you've considered**

With the current API of libhttpserver *always* cleaning up would only be possible by providing custom methods for *all* of these:

- internal_error_page
- method_not_allowed_page
- not_found_page

And you had to put that same cleanup code into *every* render method in your application, no matter if it is supposed to handle file uploads at all.

This would lead to lots of duplicated code with high potential to miss some path.

**Additional context**

Same problem applies to the draft #246 not followed (for good reasons) in favor of #257.

ctomahogh:
[...]

PR #266, opened 2022-04-07
Cleanup files uploaded

LeSpocky:
### Requirements for Contributing a Bug Fix

* [x] Fill out the template below. Any pull request that does not include enough information to be reviewed in a timely manner may be closed at the maintainers' discretion.
* [x] The pull request must only fix an existing bug. To contribute other changes, you must use a different template. You can see all templates at https://github.com/etr/libhttpserver/tree/master/.github/PULL_REQUEST_TEMPLATE.
* [x] The pull request must update the test suite to demonstrate the changed functionality.
* [x] After you create the pull request, all status checks must be pass before a maintainer reviews your contribution. For more details, please see https://github.com/etr/libhttpserver/tree/master/CONTRIBUTING.md#pull-requests.

### Identify the Bug

#264 

### Description of the Change

The list of files lives in the `httpserver::http_request` object.  The destructor is called after the request handler is completed, so we can assume user had done all tasks needed for this request.  Simply cleanup those files, before the files map is gone.

### Alternate Designs

_none considered_

This is the most simple approach.

### Possible Drawbacks

If the user needs those files, she should copy/move theme in the request
handler.

### Verification Process

Integration tests in my application, not public.

### Release Notes

"N/A" because follow-up to #257.

### DIFF
--- a/src/http_request.cpp
+++ b/src/http_request.cpp
@@ -256,4 +256,13 @@ std::ostream &operator<< (std::ostream &os, const http_request &r) {
     return os;
 }
 
+http_request::~http_request() {
+    for ( const auto &file_key : this->get_files() ) {
+        for ( const auto &files : file_key.second ) {
+            // C++17 has std::filesystem::remove()
+            remove(files.second.get_file_system_file_name().c_str());
+        }
+    }
+}
+
 }  // namespace httpserver
--- a/src/httpserver/http_request.hpp
+++ b/src/httpserver/http_request.hpp
@@ -46,6 +46,8 @@ struct MHD_Connection;
 
 namespace httpserver {
 
+namespace details { struct modded_request; }
+
 /**
  * Class representing an abstraction for an Http Request. It is used from classes using these apis to receive information through http protocol.
 **/
@@ -254,6 +256,8 @@ class http_request {
      http_request& operator=(const http_request& b) = default;
      http_request& operator=(http_request&& b) = default;
 
+     ~http_request();
+
      std::string path;
      std::string method;
      std::map<std::string, std::string, http::arg_comparator> args;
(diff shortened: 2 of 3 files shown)


LeSpocky:
Modified the integration tests to check on deleted files. Hope this is sufficient to show the mechanism works as it should.

LeSpocky:
> Modified the integration tests to check on deleted files. Hope this is sufficient to show the mechanism works as it should.

On some machines those tests fail (and sometimes with a strange `0!=0` result) and on some they don't. 

Can this happen because the webserver runs in a different thread than the test and this is some kind of concurrency problem?

etr:
Hey,

Thanks for the change.

> Not entirely sure about that friend class thing. I suppose the destructor should be private, because the constructors are also private?

Yes, it is fine with the destructor being private and using friends to access it.

> Need to come up with some way to test this in libhttpserver integration tests, right?

I guess you have the test now or is there something else you were thinking?

> Can this happen because the webserver runs in a different thread than the test and this is some kind of concurrency problem?

It might be that the check is happening before the deletion has happened. Is this still happening or your latest change has fixed it?

LeSpocky:
Thanks for looking into my changes. Answers to your questions below.

> > Need to come up with some way to test this in libhttpserver integration tests, right?
> 
> I guess you have the test now or is there something else you were thinking?

Correct. I added the changes to the tests few days after the work in the destructor.

> > Can this happen because the webserver runs in a different thread than the test and this is some kind of concurrency problem?
> 
> It might be that the check is happening before the deletion has happened. Is this still happening or your latest change has fixed it?

No, the last change to the tests fixed that. CI did complain before, but not anymore after that.

etr:
Thanks again for the help. All approved.

PR #269, opened 2022-05-13
Closes #268: Fix POST processing of final keys without values

### LINKED ISSUE #268: [BUG] Final keys without values are not processed for POST body
### Prerequisites

* [x] Put an X between the brackets on this line if you have checked that your issue isn't already filed: https://github.com/search?l=&q=repo%3Aetr%2Flibhttpserver&type=Issues

### Description

For a POST or PUT request with body format `application/x-www-form-urlencoded`, if there are final parameters with `key`, but without `value` these are left unprocessed and not being reported as arguments in the request.

### Steps to Reproduce

1 . Compile and launch this little code snippet:
```
#include <numeric>
#include <string>
#include <utility>

#include "httpserver.hpp"

using std::string;
using std::pair;

using namespace httpserver;

class hello_world_resource : public http_resource {
public:
    const std::shared_ptr<http_response> render(const http_request& req) {
        const auto& args = req.get_args();
        string recv_args = std::accumulate(args.begin(), args.end(), string {},
                [] (string& res, const pair<string,string>& arg) -> string {
                    const string key_val_str { "\"" + arg.first + "\": " + "'" + arg.second + "'" };

                    if (res.empty()) {
                        res = key_val_str;
                    } else {
                        res = res + ", " + key_val_str;
                    }

                    return res;
                }
            );

        return std::shared_ptr<http_response>(new string_response("Received args: {" + recv_args + "}"));
    }
};

int main(int argc, char** argv) {
    webserver ws = create_webserver(8080);

    hello_world_resource hwr;
    ws.register_resource("/hello", &hwr);
    ws.start(true);
    
    return 0;
}
```

2. Perform the following requests to the registered endpoint.

* curl --insecure -d "arg1=" -H "Content-Type: application/x-www-form-urlencoded" -X POST "http://127.0.0.1:8080/hello"
* curl --insecure -d "arg1=&arg2=" -H "Content-Type: application/x-www-form-urlencoded" -X POST "http://127.0.0.1:8080/hello"

**Expected behavior:**

* Response for first request **should be**: `Received args: {arg1=''}`
* Response for second request **should be**: `Received args: {arg1='', arg2=''}`

**Actual behavior:**

* Response for first request **is**: `Received args: {}`
* Response for second request **is**: `Received args: {arg1=''}`

### Versions

* OS version (if on linux, the output of "uname -a"): Linux 5.17.5-arch1-2
* libhttpserver version (please specify whether compiled or packaged): `master` compiled and 0.18.1 compiled
* libmicrohttpd version (please specify whether compiled or packaged): libmicrohttpd-0.9.68 compiled

### Additional Information

I'm submitting a PR with the potential fix.

JavierJF:
### Identify the Bug

https://github.com/etr/libhttpserver/issues/268

### Description of the Change

The change is simple, instead of calling 'MHD_destroy_post_processor' in the destructor after the request has been processed, it's called before calling the registered processing callback.

### Alternate Designs

I haven't think in alternate designs. I chose to call  'MHD_destroy_post_processor' right before `mr->callback` because I think it's the only place requiring of the final post processing of the request. Thus, keeping the two operation close makes sense.

### Possible Drawbacks

I all the information is going to be used just by `mr->callback`, can't think of any drawback.

### Verification Process

Sorry, internal testing, just testing over the endpoints in a third party application.

### Release Notes

- Fixed processing of final keys without values for POST body processing of kind 'application/x-www-form-urlencoded'


src/webserver.cpp
```diff
@@ -662,6 +662,11 @@ MHD_Result webserver::finalize_answer(MHD_Connection* connection, struct details
     if (found) {
         try {
             if (hrm->is_allowed(method)) {
+                if (mr->pp != NULL) {
```
etr, line 665: nit: can we use `nullptr` instead of `NULL`

etr (commented):
Thanks for the fix. Looks good overall - I left you a nitpicky comment about NULL vs nullptr.

I think the only missing thing is a test proving the use-case you spotted and the fact that it now works.

Issue #284, opened 2022-09-29
[BUG] use of const shared_ptr return type is ineffective/unneeded


stuart-byma:
### Prerequisites

* [x] Put an X between the brackets on this line if you have checked that your issue isn't already filed: https://github.com/search?l=&q=repo%3Aetr%2Flibhttpserver&type=Issues

### Description

This is a sort-of bug and really a minor issue. 

Using const qualifier on the shared_ptr return type is ineffective and unneeded. clang-tidy points out:

warning: return type 'const std::shared_ptr<httpserver::http_response>' is 'const'-qualified at the top level, which may reduce code readability without improving const correctness [readability-const-return-type]
std::shared_ptr<httpserver::http_response> const

See https://stackoverflow.com/questions/63972019/understanding-const-at-top-level-which-may-reduce-code-readability-without-i

for more details.

### Steps to Reproduce

To reproduce, run clang-tidy 14 or 15 on the code base, or any code that uses a `const std::shared_ptr<http_response>` return type.

**Expected behavior:** [What you expect to happen]

No issues found (at least not this one, there maybe other things that clang-tidy flags). 

**Actual behavior:** [What actually happens]

clang-tidy reports the issue. 

**Reproduces how often:** [What percentage of the time does it reproduce?]
100%

### Versions

clang-tidy 14 or 15, possibly earlier versions as well.

### Additional Information

None

etr:
It does make sense - I just approved the PR. Thanks a lot for your help.

PR #285, opened 2022-09-29
Remove ineffective/unneeded const from shared_ptr return types

stuart-byma:
### Identify the Bug

https://github.com/etr/libhttpserver/issues/284

### Description of the Change

Replace all instances of `const shared_ptr<http_response>` with `shared_ptr<http_response>`

### Alternate Designs

There is no alternate design 

### Possible Drawbacks

I don't believe there are any. 

### Verification Process

All tests pass. Everything compiles fine. clang-tidy no longer reports the issue in a code base that uses this return type.

### Release Notes

Remove ineffective/unneeded const from std::shared_ptr <http_response> return types.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ The most basic example of creating a server and handling a requests for the path
 
     class hello_world_resource : public http_resource {
     public:
-        const std::shared_ptr<http_response> render(const http_request&) {
+        std::shared_ptr<http_response> render(const http_request&) {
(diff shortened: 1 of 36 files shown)


etr:
Thanks for the cleanup.

jcphill:
I know it's a bit late but this should have been flagged as a breaking interface change because existing code that overrides httpserver::http_resource::render_GET etc. now has a "invalid covariant return type" error and there is no way to write code that works with both the old const and new non-const return types.

Issue #286, opened 2022-10-28
move toward a more zero-copy API


stuart-byma:
**Is your feature request related to a problem? Please describe.**
Currently when getting any kind of header, footer, cookie, etc. new string objects are constructed every time. This leads to a large number of memory allocations, which are generally expensive. Even if I have a key value as a `const char*` or a `string_view` I am forced to construct a std::string to use the interface. Even worse is requesting all keys and values, where a std::map is allocated and constructed and every key and value copied into a std::string. 

**Describe why the feature or enhancement you are proposing fits the library.**
Performance should be a priority for any HTTP server :smile:  

**Describe the solution you'd like**
The API calls should favor `string_view` or similar to pass views on underlying data in MHD (which is mostly immutable) and avoid having to make copies. Any MHD data should be viable for the lifetime of the HTTP request object anyway. If the user really wants a copy to keep the data around longer, they can make a copy explicitly themselves.

**Describe alternatives you've considered**
I'm not sure there is any better option than using `std::string_view` or similar here.

**Additional context**
None.

PR #287, opened 2022-10-28
begin move toward more zero copy API

stuart-byma:
### Requirements for Adding, Changing, or Removing a Feature

* Fill out the template below. Any pull request that does not include enough information to be reviewed in a timely manner may be closed at the maintainers' discretion.
* The pull request must contribute a change that has been endorsed by the maintainer team. See details in the template below.
* The pull request must update the test suite to exercise the updated functionality.
* After you create the pull request, all status checks must be pass before a maintainer reviews your contribution. For more details, please see https://github.com/etr/libhttpserver/tree/master/CONTRIBUTING.md#pull-requests.

https://github.com/etr/libhttpserver/issues/286

### Description of the Change

Update main APIs in the HTTPRequest type to use std::string_view instead of std::string. This eliminates many string copies and unnecessary memory allocations by simple returning / passing views over immutable string data.

### Alternate Designs

No other alternatives. The change is relatively simple.

### Possible Drawbacks

This is a potentially breaking change to the API. Users may have to update their apps after updating. 

### Verification Process

All tests passing. Added additional testing for new API interfaces. 

### Release Notes

Many instances of `std::string` within the API changed to `std::string_view` to improve performance by reducing copies and memory allocations.

### DIFF
--- a/src/httpserver/http_request.hpp
+++ b/src/httpserver/http_request.hpp
@@ -118,11 +118,23 @@ class http_request {
      const std::map<std::string, std::string, http::header_comparator> get_headers() const;
 
      /**
+      * Method used to get all header key values passed with the request.
+      * @param keys a vector of string_view into which the keys will be inserted
+     **/
+     void get_header_keys(std::vector<std::string_view>* keys) const;
+
+     /**
       * Method used to get all footers passed with the request.
       * @param result a map<string, string> > that will be filled with all footers
       * @result the size of the map
      **/
      const std::map<std::string, std::string, http::header_comparator> get_footers() const;
+
+     /**
+      * Method used to get all footer key values passed with the request.
+      * @param keys a vector of string_view into which the keys will be inserted
+     **/
+     void get_footer_keys(std::vector<std::string_view>* keys) const;
 
      /**
       * Method used to get all cookies passed with the request.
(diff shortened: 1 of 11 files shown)


stuart-byma:
It looks like the autobuild for this PR is now failing purely due to some configs using ancient compiler versions that are pre-c++17. Any chance we can just nuke those configs? Pretty sure Ubuntu 18.04 is also end of life.

etr:
We can update the config here: https://github.com/etr/libhttpserver/blob/master/.github/workflows/verify-build.yml.

Worth removing old versions of the compilers, and adding the new versions of the same. OS-wise, if the compilers are not present, we can bump it up to the minimum version that has them (e.g. 20.04)

src/httpserver/http_request.hpp
```diff
@@ -117,27 +117,51 @@ class http_request {
      **/
      const std::map<std::string, std::string, http::header_comparator> get_headers() const;
 
+     /**
+      * Method used to get all header key values passed with the request.
+      * @param keys a vector of string_view into which the keys will be inserted
+     **/
+     void get_header_keys(std::vector<std::string_view>* keys) const;
```
etr, line 124: Why are we bundling these new public methods into the zero-copy change?

stuart-byma: The rationale is that you often need to know what keys are present before getting them. The only option otherwise is to the use the "get_headers" which returns everything, also incurring the cost of constructing/allocating the map. We could also change the "get_headers" method and friends to returns maps of string_view as well, but that still incurs the cost of constructing the map. I can add that to the PR if you want. Let me know what you prefer.

etr: Having methods to fetch the keys might be fine, but it should be isolated in its own PR not part of this one, so we have the PR doing one job only.

src/httpserver/http_response.hpp
```diff
@@ -81,6 +81,10 @@ class http_response {
          return cookies[key];
      }
 
+     std::string_view get_content_type() const {
+         return headers.at(http::http_utils::http_header_content_type).c_str();
+     }
+
```
etr, line 87: Same comment as above - not sure this fits the context and purpose of the PR

src/httpserver/string_response.hpp
```diff
@@ -55,6 +55,7 @@ class string_response : public http_response {
 
      MHD_Response* get_raw_response();
 
+     std::string_view get_content() const { return std::string_view(content.c_str()); }
```
etr, line 58: Here as well, I don't think this should be part of the PR.

stuart-byma:
I'm going to resubmit a new PR for this issue that takes comments from @etr into account and will have a cleaner commit history.

PR #288, opened 2022-10-29
Issue #286: use string_view in API methods where applicable

stuart-byma:

https://github.com/etr/libhttpserver/issues/286

### Description of the Change

This change reduces required copying of immutable data primarily in the HTTP request API. Methods returning strings now return `string_view` to avoid copying immutable data from the underlying MHD impl. `get_*s` methods still return a `std::map` which must be allocated and constructed, but does not copy underlying strings.

Fix a few instances of these method calls in the code base.

Introduce some type aliases for maps from `string_view` to `string_view`, `string` to `string` to reduce code verbosity.
Overload the comparator `operator()`s as required.

Provide overloads for `dump_*_map` functions and a template impl to avoid code duplication.

Bump C++ version to 17 required to use `string_view`. Remove CI configurations using ancient compilers that don't support `string_view` or C++17

### Alternate Designs

No other alternatives. The change is relatively simple.

### Possible Drawbacks

This is a potentially breaking change to the API. Users may have to update their apps after updating. 

### Verification Process

All tests passing. 

### Release Notes

Many instances of `std::string` within the API changed to `std::string_view` to improve performance by reducing copies and memory allocations.


src/httpserver/http_utils.hpp
```diff
@@ -301,6 +304,13 @@ class arg_comparator {
...
       * @param second string
      **/
+     bool operator()(std::string_view x, std::string_view y) const {
+#ifdef CASE_INSENSITIVE
+         COMPARATOR(x, y, std::toupper);
+#else
+         COMPARATOR(x, y,);  // NOLINT(whitespace/comma)
+#endif
+     }
      bool operator()(const std::string& x, const std::string& y) const {
```
etr, line 314: Given we have the "string_view" version, I'd have the "string" version just use the string view one (which saves an ifdef/else block)

src/httpserver/http_utils.hpp
```diff
@@ -356,15 +371,17 @@ uint16_t get_port(const struct sockaddr* sa);
  * @param prefix Prefix to identify the map
  * @param map
 **/
-void dump_header_map(std::ostream &os, const std::string &prefix, const std::map<std::string, std::string, header_comparator> &map);
+void dump_header_map(std::ostream &os, const std::string &prefix, const http::value_map &map);
+void dump_header_map(std::ostream &os, const std::string &prefix, const http::value_map_view &map);
```
etr, line 375: I would go one way or the other for this dump_* methods and not keep both versions. These are mostly for debug purposes.

src/webserver.cpp
```diff
@@ -463,7 +463,7 @@ MHD_Result webserver::post_iterator(void *cls, enum MHD_ValueKind kind,
 
     try {
         if (filename == nullptr || mr->ws->file_upload_target != FILE_UPLOAD_DISK_ONLY) {
-            mr->dhr->set_arg(key, mr->dhr->get_arg(key) + std::string(data, size));
+            mr->dhr->set_arg(key, std::string(mr->dhr->get_arg(key)) + std::string(data, size));
```
etr, line 466: Why do we need the explicit cast here? Doesn't string_view have a conversion constructor that does it transparently?  (asking for the various places where this happens, not only this instance)

src/httpserver/http_utils.hpp
```diff
@@ -310,6 +320,11 @@ class arg_comparator {
      }
 };
 
+using value_map = std::map<std::string, std::string, http::header_comparator>;
+using arg_map = std::map<std::string, std::string, http::arg_comparator>;
+using value_map_view = std::map<std::string_view, std::string_view, http::header_comparator>;
+using arg_map_view = std::map<std::string_view, std::string_view, http::arg_comparator>;
```
etr, line 326: I am not sure I'd keep these - the maps aren't used enough around that it is worth it.

etr (commented):
Mostly minor comments from me - sorry for the delay.

src/webserver.cpp
stuart-byma: Yes, and nope: https://stackoverflow.com/questions/47525238/why-is-there-no-implicit-conversion-from-stdstring-view-to-stdstring

src/httpserver/http_utils.hpp
stuart-byma: I would agree, but the problem is that the HttpResponse type owns its header/footer data and therefore has `value_map`s and not `value_map_view`s. So there are places where dump_header_map is used on both owning and non-owning map types.

src/httpserver/http_utils.hpp
stuart-byma: Good idea :+1:

src/httpserver/http_utils.hpp
stuart-byma: There are ~52 instances of these types throughout the codebase, I typically use aliases for less. These kinds of template decls are super verbose as well :smile:

etr: That's a lot more than I thought. Fair to keep them - I'll open an issue for a general cleanup

src/httpserver/http_utils.hpp
etr: I wouldn't bother trying to produce a "view" version of this one as it is a method used for debugging only. Internally to the library only http_response and http_request use it (both in their internal operations) and a test for http_utils that builds the map as map of strings. I think just keeping the old ones here is fine - outside of this PR, I'll move this under "details" to make it clear that it is not something we offer as interface, anyway.

src/webserver.cpp
etr: makes sense

src/httpserver/http_utils.hpp
etr: Just fix the names for clarity:

- value_map -> header_map
- value_map_view -> header_view_map
- arg_map is fine
- arg_map_view -> arg_view_map

src/http_request.cpp
```diff
@@ -114,8 +114,8 @@ const std::string http_request::get_arg(const std::string& key) const {
     return get_connection_value(key, MHD_GET_ARGUMENT_KIND);
 }
 
-const std::map<std::string, std::string, http::arg_comparator> http_request::get_args() const {
-    std::map<std::string, std::string, http::arg_comparator> arguments;
+const http::arg_map http_request::get_args() const {
```
etr, line 117: would it make sense for this to return a map of view as for the rest?

README.md
```diff
@@ -559,14 +559,14 @@ The `http_request` class has a set of methods you will have access to when imple
...
+* _**const std::map<std::string_view, std::string_view, http::header_comparator>** get_headers() **const**:_ Returns a map containing all the headers present in the HTTP request.
+* _**const std::map<std::string_view, std::string_view, http::header_comparator>** get_cookies() **const**:_ Returns a map containing all the cookies present in the HTTP request.
+* _**const std::map<std::string_view, std::string_view, http::header_comparator>** get_footers() **const**:_ Returns a map containing all the footers present in the HTTP request (only for http 1.1 chunked encodings).
+* _**const std::map<std::string_view, std::string_view, http::arg_comparator>** get_args() **const**:_ Returns all the arguments present in the HTTP request. Arguments can be (1) querystring parameters, (2) path argument (in case of parametric endpoint, (3) parameters parsed from the HTTP request body if the body is in `application/x-www-form-urlencoded` or `multipart/form-data` formats and the postprocessor is enabled in the webserver (enabled by default).
```
etr, line 569: This doesn't match the implementation below (which still seems not to be using views)

stuart-byma: Updated. I think I caught all instances, but if you can double check that would be great.

src/httpserver/http_utils.hpp
stuart-byma: I actually left just the view version. The test case usages are switched to view maps (trivial) which left HttpResponse as the only place data owning versions were being dumped. I added utility to convert the map to a view just for that case. I think this results in the fewest lines of code changed. The utility (to_view_map) can also just go in the details namespace when you make that change.

src/http_request.cpp
stuart-byma: Actually ran into an issue with this. `build_request_args` actually makes a temp copy of the underlying data and does some unescape stuff on it. We cannot return a view over that temp data since it goes out of scope.

etr: uhm...that is a really annoying inconsistency in the interface that will leave folks puzzled. I need to think of whether there is a way out (philosophically, I'd rather prioritize clarity/consistency over performance, unless major)

etr: To be fair, all of this originates from some old "unescaping" logic that made us bypass the mhd unescaper and build a separated unescaping logic on top of it. I think we might have to unblock that first (and I think it should be possible, given that nobody has touched that code in the past 10 years and existed to fix some misbehavior that was there in a dependency at that time).

src/httpserver/http_request.hpp
```diff
@@ -159,29 +159,29 @@ class http_request {
       * @param key the specific header to get the value from
       * @return the value of the header.
      **/
-     const std::string get_header(const std::string& key) const;
+     std::string_view get_header(std::string_view key) const;
```
etr, line 162: Should we align the rest of the interface of this class to use string_view? (e.g. get_user, get_pass, get_method, etc.)

src/http_request.cpp
stuart-byma: Agree that can be handled as a separate issue. Afterward we can make get_args return an arg_view_map.

stuart-byma: Do we think the MHD unescaper is fine at this point? Maybe worth checking.

stuart-byma: get_args firsts takes a copy of the args member var, then goes to MHD to get additional args. Can we use the args member var to just cache the unescaped values from MHD, and then return an arg_view_map constructed from the args member var?

stuart-byma: Sorry for the barrage of comments, but I notice get_arg to get a single arg does not do any unescaping, why is that?

src/httpserver/http_request.hpp
stuart-byma: I think that would be best. Design decisions to make though: some of these other methods modify data extracted from MHD (e.g. get_querystring), construct a result and then return it. I see two options here. One is to change the signature and/or return values to match whatever MHD gives, and return views over that data. The other is to construct the result as now only once, cache in the request object, and return a view over that (also if the user calls twice, the cached value is simply returned on the second call). Thoughts?

etr: It's unfortunate that I am not sure I really like any of the five options I see for this. We basically can:

1. Leave the logic as is, and have an inconsistent interface. Some methods will return string, others return string_view. The issue with it is that basically we are basically projecting outside our own internal logic and the user pays the cost for this by having to adapt to what each method returns for a reason that is not immediately obvious when in their position.

2. Revert the methods to pure proxies over what libmicrohttpd returns. This comes at a loss of the extra simplicity that httpserver provides (which is one of the core reasons people use this library).

3. Pre-fetch all data from libmicrohttpd and store it preventively in the request object. This would allow to use string_view everywhere. This used to be the old implementation, but we moved away from it because it was just way too expensive for no reason (the user pays the cost of the fetch + copy even when they never use the data).

4. Do a fetch and cache. This is an implementation I had thought of when migrating off of the old prefetch approach. It has the extra advantage of making subsequent calls cheaper. It has a massive issue that it makes all getters non const which is a really annoying property for the interface of what basically is a class meant to store and transport data. I considered and discarded this approach when moving away from the old pre-fetch one.

5. Just leave the class as is and not use string_view at all.

I am currently pondering 1 an 5 - everything else just doesn't seem worth it. Instinctively, I'd say 5 would be my general inclination because philosophically I'd rather prioritize clarity/consistency over marginal performance, but I can welcome disagreements, if well motivated. I also think it would be worth to simulate some heavy data fetch (e.g. of headers) with both approaches to actually prove that using string_view whenever possible is worth the inconsistency it costs interface-wise.

src/http_request.cpp
etr: > Sorry for the barrage of comments, but I notice get_arg to get a single arg does not do any unescaping, why is that?

It sounds very much like a bug. To be fair, I can see there is some old code in here that makes no sense anymore (see it tries to get data from an "args" dataset that never gets filled because the method is const). I think the general management of args and unescaping has issues that need fixing.

src/httpserver/http_request.hpp
stuart-byma: I agree that 3 is untenable performance-wise. Most users are not going to use every field, prefetch has a high init cost. 1 is also undesirable; I agree it is better to be as consistent as possible.

I am not sure I agree with your assessment of point 4. An object internal cache doesn't mean we have to sacrifice `const` on all the accessors. It just means we can't cache the items directly in the request object. We can use an private `cache` type, and store it within a `unique_ptr`. Then we have a mutable cache from within a `const` accessor. There are only a few values that need to be cached in this way for us to make the whole API return `string_view`. 

I think I prefer this option, it seems to be the best compromise between API consistency, const-ness, and performance. I have already tried this out, I will add it to the PR and look forward to your thoughts.

src/httpserver/http_request.hpp
```diff
@@ -240,27 +241,27 @@ class http_request {
...
-     http_request& operator=(const http_request& b) = default;
+     http_request& operator=(const http_request& b) = delete;
      http_request& operator=(http_request&& b) = default;
 
      ~http_request();
 
      std::string path;
      std::string method;
-     std::map<std::string, std::string, http::arg_comparator> args;
+     http::arg_map args;
```
etr, line 264: We should just remove this, as it isn't used and just muddles the waters on where the cache really is.

src/httpserver/http_request.hpp
```diff
@@ -240,27 +241,27 @@ class http_request {
...
-      * @param b http_request b to copy attributes from.
+      * Move constructor.
+      * @param b http_request b to move attributes from.
      **/
-     http_request(const http_request& b) = default;
      http_request(http_request&& b) noexcept = default;
 
-     http_request& operator=(const http_request& b) = default;
+     http_request& operator=(const http_request& b) = delete;
      http_request& operator=(http_request&& b) = default;
```
etr, line 258: It's worth adding a piece of in line documentation for these two lines to explain why we are doing what we are doing here. I think it can become confusing easily for people that come and read this without context.

src/httpserver/http_request.hpp
```diff
@@ -356,8 +359,23 @@ class http_request {
...
      }
 
-     const std::string get_connection_value(const std::string& key, enum MHD_ValueKind kind) const;
-     const std::map<std::string, std::string, http::header_comparator> get_headerlike_values(enum MHD_ValueKind kind) const;
+     // Cache certain data items on demand so we can consistently return views
+     // over the data. Some things we transform before returning to the user for
+     // simplicity (e.g. query_str, requestor), others out of necessity (arg unescaping).
+     // Others (username, password, digested_user) MHD returns as char* that we need
+     // to make a copy of and free anyway.
+     struct data_cache {
```
etr, line 367: maybe rename to http_request_data_cache which is more explicit

src/httpserver/http_request.hpp
```diff
@@ -356,8 +359,23 @@ class http_request {
...
+        std::string username;
+        std::string password;
+        std::string query_str;
+        std::string requestor_ip;
+        std::string digested_user;
+        http::arg_map unescaped_mhd_args;
```
etr, line 373: I'd just call it unescaped_args (mhd just happens to be our current underlying framework, but we aren't necessarily bound to it)

src/httpserver/http_request.hpp
```diff
@@ -276,6 +277,8 @@ class http_request {
 
      static MHD_Result build_request_querystring(void *cls, enum MHD_ValueKind kind, const char *key, const char *value);
 
+    void fetch_user_pass() const;
+
```
etr, line 281: This should be a private method

src/httpserver/http_request.hpp
```diff
@@ -276,6 +277,8 @@ class http_request {
 
      static MHD_Result build_request_querystring(void *cls, enum MHD_ValueKind kind, const char *key, const char *value);
 
+    void fetch_user_pass() const;
```
etr, line 280: nit: indent by an additional character

src/httpserver/http_request.hpp
etr, line 367: Just move this at the bottom, just before the friends

src/httpserver/http_request.hpp
etr: That's a good idea - I am looking at the implementation and it seems like it lands in a good place. I've just left a few minor comments. Take a look at those. I'll take another re-read later to give you a more complete feedback.

src/http_request.cpp
```diff
@@ -32,17 +32,17 @@ const char http_request::EMPTY[] = "";
...
+    http::arg_map* arguments;
 };
 
 void http_request::set_method(const std::string& method) {
     this->method = string_utilities::to_upper_copy(method);
 }
 
 bool http_request::check_digest_auth(const std::string& realm, const std::string& password, int nonce_timeout, bool* reload_nonce) const {
-    std::string digested_user = get_digested_user();
+    auto digested_user = get_digested_user();
```
etr, line 43: Let's keep using std::string here for clarity

src/http_request.cpp
```diff
@@ -173,47 +181,50 @@ MHD_Result http_request::build_request_querystring(void *cls, enum MHD_ValueKind
...
 
-const std::string http_request::get_digested_user() const {
-    char* digested_user_c = nullptr;
-    digested_user_c = MHD_digest_auth_get_username(underlying_connection);
+std::string_view http_request::get_digested_user() const {
+    if (!cache->digested_user.empty()) {
+        return cache->digested_user;
+    }
+
+    auto* digested_user_c = MHD_digest_auth_get_username(underlying_connection);
```
etr, line 219: I would use char* for clarity here

src/http_response.cpp
```diff
@@ -54,12 +54,22 @@ void http_response::shoutCAST() {
     response_code |= http::http_utils::shoutcast_response;
 }
 
+namespace {
+static inline http::header_view_map to_view_map(const http::header_map & hdr_map) {
```
etr, line 58: nit: keep '&' attached on the left

src/http_response.cpp
```diff
@@ -54,12 +54,22 @@ void http_response::shoutCAST() {
...
+    http::header_view_map view_map;
+    for (const auto & item : hdr_map) {
```
etr, line 60: nit: keep '&' attached on the left

src/http_utils.cpp
```diff
@@ -523,17 +524,12 @@ void dump_header_map(std::ostream &os, const std::string &prefix, const std::map
...
 
-    if (map.size()) {
-        os << "    " << prefix << " [";
-        for (; it != end; ++it) {
-            os << (*it).first << ":\"" << (*it).second << "\" ";
-        }
-        os << "]" << std::endl;
-    }
+void dump_arg_map(std::ostream &os, const std::string &prefix, const http::arg_view_map &map) {
+    dump_map<decltype(map)>(os, prefix, map);
```
etr, line 532: use arg_view_map instread of decltype here; it doesn't buy much

src/httpserver/http_request.hpp
```diff
@@ -240,27 +241,32 @@ class http_request {
...
 
      http_request(MHD_Connection* underlying_connection, unescaper_ptr unescaper):
          underlying_connection(underlying_connection),
-         unescaper(unescaper) { }
+         unescaper(unescaper), cache(std::make_unique<http_request_data_cache>()) {}
 
      /**
-      * Copy constructor.
-      * @param b http_request b to copy attributes from.
+      * Copy constructor. Deleted to make class move-only. Also required due to presence of unique_ptr member.
```
etr, line 251: I would be more verbose on why we want the object to be move only

etr:
Only minor comments on the change at this point - I think once we close those, we are done.

src/http_response.cpp
etr: not sure you did this?

etr: I noticed others below - can you make sure you keep the reference '&' consistently together (left attached) to the type in all of your changes?

I am aware the code is inconsistent (even in some of the rows you are modifying), but let's make sure we fix everything we are touching or creating anew. I need to go and have a pass at cleaning everything up later for the rest..

stuart-byma: I think I got them all now. I would suggest configuring something clang-format, though the initial application of it will likely make quite a few changes that are formatting-only.

etr: Yep, there is a change pending on that (with a few thousands fixes) that I am going through.

etr:
Thanks for going through with this. It is merged now!

PR #293, opened 2022-12-22
fix #292 change get_args to return mutliple values per key

### LINKED ISSUE #292: [BUG] multiple args with the same key are not handled correctly
### Prerequisites

* [X ] Put an X between the brackets on this line if you have checked that your issue isn't already filed: https://github.com/search?l=&q=repo%3Aetr%2Flibhttpserver&type=Issues

### Description

A query arg string can have multiple arguments with the same key, e.g. `hello.com/what?test=one&test=two`. libhttpserver only ever returns the last arg via get_args(). This is inconsistent with MHD, which will iterate multiple args with the same key. get_args() simply overwrites the last one. 

This is also inconsistent with other http library, such as Go http package, which will return a collection of strings for a given argument key.

### Steps to Reproduce

**Expected behavior:** get_args() returns a map contains all the values associated to one key, even if there are multiple.

**Actual behavior:** get_args() overwrites and always returns the last value for given key.

### Versions

Not affected by versions.

stuart-byma:
### Requirements for Contributing a Bug Fix

* Fill out the template below. Any pull request that does not include enough information to be reviewed in a timely manner may be closed at the maintainers' discretion.
* The pull request must only fix an existing bug. To contribute other changes, you must use a different template. You can see all templates at https://github.com/etr/libhttpserver/tree/master/.github/PULL_REQUEST_TEMPLATE.
* The pull request must update the test suite to demonstrate the changed functionality.
* After you create the pull request, all status checks must be pass before a maintainer reviews your contribution. For more details, please see https://github.com/etr/libhttpserver/tree/master/CONTRIBUTING.md#pull-requests.

### Identify the Bug

https://github.com/etr/libhttpserver/issues/292

### Description of the Change

Change arg_map and arg_view_map to `std::map<std::string, std::vector<std::string>>` etc, so that multiple values per key can be returned.

Update an appropriate test case to check that the behavior is correct.

### Alternate Designs

No alternate designs

### Possible Drawbacks

Small change to user facing API.

### Verification Process

Tests were updated to check that new code is correct.

### Release Notes

 - Fixed a bug where http_request::get_args() only returned the last value for a given argument key. It now returns all values associated with a given key.

### DIFF
--- a/src/http_request.cpp
+++ b/src/http_request.cpp
@@ -105,20 +105,31 @@ const http::header_view_map http_request::get_cookies() const {
 }
 
 std::string_view http_request::get_arg(std::string_view key) const {
-    std::map<std::string, std::string>::const_iterator it = cache->unescaped_args.find(std::string(key));
+    auto const it = cache->unescaped_args.find(std::string(key));
 
     if (it != cache->unescaped_args.end()) {
-        return it->second;
+        return it->second.front();
     }
 
     return get_connection_value(key, MHD_GET_ARGUMENT_KIND);
+}
+
+static void fill_arg_view_map(http::arg_view_map *arguments, const http::arg_map& cached_map) {
+    for (const auto& kv : cached_map) {
+        std::vector<std::string_view> vec;
+        vec.reserve(kv.second.size());
+        for (const auto& value : kv.second) {
+            vec.push_back(value);
+        }
+        arguments->emplace(std::string_view(kv.first), std::move(vec));
+    }
 }
 
 const http::arg_view_map http_request::get_args() const {
     http::arg_view_map arguments;
 
     if (!cache->unescaped_args.empty()) {
-        arguments.insert(cache->unescaped_args.begin(), cache->unescaped_args.end());
+        fill_arg_view_map(&arguments, cache->unescaped_args);
         return arguments;
     }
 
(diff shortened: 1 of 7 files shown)


src/http_request.cpp
```diff
@@ -155,7 +166,13 @@ MHD_Result http_request::build_request_args(void *cls, enum MHD_ValueKind kind,
     std::string value = ((arg_value == nullptr) ? "" : arg_value);
 
     http::base_unescaper(&value, aa->unescaper);
-    (*aa->arguments)[key] = value;
+    auto existing_iter = aa->arguments->find(key);
+    if (existing_iter != aa->arguments->end()) {
+        // Key exists, add value to collection instead of overwriting previous value.
```
mdf-observe, line 171: Whether the key exists or not, I think this if/else equivalent to just:

    (*aa->arguments)[key].push_back(value);

src/httpserver/http_request.hpp
```diff
@@ -298,7 +298,12 @@ class http_request {
       * @param value The value assumed by the argument
      **/
      void set_arg(const std::string& key, const std::string& value) {
-         cache->unescaped_args[key] = value.substr(0, content_size_limit);
+         auto existing_iter = cache->unescaped_args.find(key);
+         if (existing_iter != cache->unescaped_args.end()) {
```
mdf-observe, line 302: Similarly I think this is also just `cache->unescaped_args[key].push_back(value.substr(0, content_size_limit));`, and below.

src/http_request.cpp
```diff
@@ -105,20 +105,31 @@ const http::header_view_map http_request::get_cookies() const {
 }
 
 std::string_view http_request::get_arg(std::string_view key) const {
```
mdf-observe, line 107: Since get_arg() now returns only the first matching item (which does keep the API the same), should there also be a get_args() method that returns a `std::span<const std::string>` or whatever type makes the least memory allocations?  IIRC a `span<const T>` can be made from a `vector<T>` implicitly.

stuart-byma: afaik span is C++20, so this would require bumping the min version from 17.

PR #299, opened 2023-02-15
Fix a typo (my bad) in the docs.

bcsgh:
### Description of the Change

The wrong name is used for one of the documented methods.

### Release Notes

Fix a typo in the docs.


etr:
Thanks for the fix

PR #302, opened 2023-03-05
fix #292 handle argument keys with multiple values

### LINKED ISSUE #292: [BUG] multiple args with the same key are not handled correctly

stuart-byma:
### Identify the Bug

#292 

### Description of the Change

This change allows the server to handle argument keys with multiple values e.g. url?arg=param&arg=what

Previous, the arg key would just be overwritten with the next value. This change adds a http_arg_value type to house all the values for a given key.

Adjust existing get_arg(s) methods, and add some get_arg_flat type methods that still just return the first value for a given key.

Add a test case for multiple valued keys.

Add a test case for a large file upload that triggers chunking behavior of MHD. This causes the server to concatenate file chunks within the first value of the given arg key, so the test ensures that this behavior is not affected by the changes.

### Alternate Designs

Design is according to discussion in #292 

### Possible Drawbacks

Small changes to public API.

### Verification Process

Additional testing added as in change description. In particular, verify that large file handling behavior has not regressed.

### Release Notes

- Multiple valued arguments are now supported, e.g. URLs containing ?arg=param1&arg=param2. `get_arg()` now returns an `http_arg_value` type containing a collection of all values associated with the requested key. `get_arg_flat()` retains the old behavior where just one value is selected and returned.

### DIFF
--- a/examples/hello_world.cpp
+++ b/examples/hello_world.cpp
@@ -33,7 +33,7 @@ class hello_world_resource : public httpserver::http_resource {
 std::shared_ptr<httpserver::http_response> hello_world_resource::render(const httpserver::http_request& req) {
     // It is possible to store data inside the resource object that can be altered through the requests
     std::cout << "Data was: " << data << std::endl;
-    std::string_view datapar = req.get_arg("data");
+    std::string_view datapar = req.get_arg_flat("data");
     set_some_data(datapar == "" ? "no data passed!!!" : std::string(datapar));
     std::cout << "Now data is:" << data << std::endl;
 
--- a/src/http_request.cpp
+++ b/src/http_request.cpp
@@ -32,7 +32,7 @@ const char http_request::EMPTY[] = "";
 
 struct arguments_accumulator {
     unescaper_ptr unescaper;
-    http::arg_map* arguments;
+    std::map<std::string, std::vector<std::string>, http::arg_comparator>* arguments;
 };
 
 void http_request::set_method(const std::string& method) {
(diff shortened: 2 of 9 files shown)


src/httpserver/http_utils.hpp
```diff
@@ -316,10 +316,36 @@ class arg_comparator {
      }
 };
 
+class http_arg_value {
```
etr, line 319: I think it would make sense to move this in its own hpp and cpp reserved files

src/http_request.cpp
```diff
@@ -104,35 +104,63 @@ const http::header_view_map http_request::get_cookies() const {
...
+    arguments_accumulator aa;
+    aa.unescaper = unescaper;
+    aa.arguments = &cache->unescaped_args;
+    MHD_get_connection_values(underlying_connection, MHD_GET_ARGUMENT_KIND, &build_request_args, reinterpret_cast<void*>(&aa));
+}
+
+http::http_arg_value http_request::get_arg(std::string_view key) const {
+    populate_args();
 
+    auto it = cache->unescaped_args.find(std::string(key));
```
mdf-observe, line 120: can we have a std::string_view comparator in http::arg_comparator to avoid this allocation?

examples/hello_world.cpp
```diff
@@ -33,7 +33,7 @@ class hello_world_resource : public httpserver::http_resource {
 std::shared_ptr<httpserver::http_response> hello_world_resource::render(const httpserver::http_request& req) {
     // It is possible to store data inside the resource object that can be altered through the requests
     std::cout << "Data was: " << data << std::endl;
-    std::string_view datapar = req.get_arg("data");
+    std::string_view datapar = req.get_arg_flat("data");
```
etr, line 36: Do you think it would be worth having an operator for string_view in http_arg_value?

test/integ/basic.cpp
```diff
@@ -627,6 +645,30 @@ LT_BEGIN_AUTO_TEST(basic_suite, postprocessor)
...
+    CURL *curl = curl_easy_init();
+    CURLcode res;
+    curl_easy_setopt(curl, CURLOPT_URL, "localhost:8080/base?arg=beep&arg=boop&arg=hello&arg=what");
+    curl_easy_setopt(curl, CURLOPT_POSTFIELDS, "arg=beep&arg=boop&arg=hello&arg=what");
+    curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, writefunc);
+    curl_easy_setopt(curl, CURLOPT_WRITEDATA, &s);
+    struct curl_slist *list = NULL;
+    // Change the content type to trigger "normal" POST processing,
+    // otherwise the file processing logic is triggered which does
+    // not set the multiple arg values as expected.
```
etr, line 662: I think you might have lost me here. Why does this need to be done?

src/webserver.cpp
```diff
@@ -463,7 +463,7 @@ MHD_Result webserver::post_iterator(void *cls, enum MHD_ValueKind kind,
 
     try {
         if (filename == nullptr || mr->ws->file_upload_target != FILE_UPLOAD_DISK_ONLY) {
-            mr->dhr->set_arg(key, std::string(mr->dhr->get_arg(key)) + std::string(data, size));
+            mr->dhr->set_arg_flat(key, std::string(mr->dhr->get_arg(key)) + std::string(data, size));
         }
```
etr, line 467: I think we should split these two use cases. If filename is null, it means it is not a file we are processing, so we can manage the normal flow with multiple arguments. The only case where we want to keep appending is the case where you have a filename && file_upload_target is not FILE_UPLOAD_DISK_ONLY (so, basically, this argument is a file argument and we are trying to load the file into memory)

test/integ/basic.cpp
stuart-byma: Without a different content type the defaults that cURL sets trigger the file processing code, but in this test I am just trying to check the "normal" arg handling with multiple values.

src/http_request.cpp
stuart-byma: The arg_comparator supports string_view, but std::map and friends do not support heterogeneous lookup? Switching the unescaped args to be keyed by string_view breaks the file handling, because the keys end up spread across several requests and become invalidated.

mdf-observe: Since C++14 std::map supports lookup by an arbitrary key type `K` in `find()`

src/webserver.cpp
stuart-byma: Maybe a bit orthogonal to this change? I can file a separate issue.

src/http_request.cpp
stuart-byma: Just had to rejig the comparator a bit. Thanks!

examples/hello_world.cpp
stuart-byma: Done

stuart-byma:
Some auto checks were cancelled? Never seen that before!

src/webserver.cpp
etr: I think it is the same issue, because we kept the behaviors of URL style args parsing consistent across GET and POST. Without the above, they'd behave inconsistently.

etr:
> Some auto checks were cancelled? Never seen that before!

I suspect that github has finally deprecated ubuntu 18.04 - I've started noticing this issue elsewhere as well. I think we can ignore that check for the purpose of this PR.

src/httpserver/http_utils.hpp
```diff
@@ -299,6 +301,7 @@ class header_comparator {
 **/
 class arg_comparator {
  public:
+     using is_transparent = std::true_type;
```
etr, line 304: What is this doing? (I cannot find it used elsewhere)

test/integ/file_upload.cpp
```diff
@@ -470,6 +513,43 @@ LT_BEGIN_AUTO_TEST(file_upload_suite, file_upload_memory_only_incl_content)
     delete ws;
 LT_END_AUTO_TEST(file_upload_memory_only_incl_content)
 
+LT_BEGIN_AUTO_TEST(file_upload_suite, file_upload_large_content)
```
etr, line 516: I think we might also benefit from a test that combines this with arguments being passed with the POST

src/httpserver/http_utils.hpp
stuart-byma: This is required to activate heterogeneous lookup using this comparator. `is_transparent` just needs to be defined, it doesn't get referenced anywhere. From cppreference:

The member type is_transparent indicates to the caller that this function object is a transparent function object: it accepts arguments of arbitrary types and uses perfect forwarding, which avoids unnecessary copying and conversion when the function object is used in heterogeneous context, or with rvalue arguments. In particular, template functions such as [std::set::find](https://en.cppreference.com/w/cpp/container/set/find) and [std::set::lower_bound](https://en.cppreference.com/w/cpp/container/set/lower_bound) make use of this member type on their Compare types.

etr:
Thanks for the change and for the patience to go through the review process. I think this is a great feature to have in the library.

Issue #305, opened 2023-04-02
[BUG] register_resource won't accept the same resource string for both family=true and family=false.


bcsgh:
### Prerequisites

* [x] Put an X between the brackets on this line if you have checked that your issue isn't already filed: https://github.com/search?l=&q=repo%3Aetr%2Flibhttpserver&type=Issues

### Description

The following code will fail

```
void RegisterResources(httpserver::webserver *ws) {
  CHECK(ws->register_resource("/", root_page, false));
  CHECK(ws->register_resource("/", fallback, true));  // This fails
}
```

In general, only the first register resource for any given path will be accepted, even if the resources would match differently.

The intent in this specific case is to have the second handler as a fallback that deals with anything not otherwise handled (e.g. return a 401 if not logged in or 404 otherwise). At the same time, there is still a need to handle the root URL which incidentally uses the same resource string. 

I know it's possible to handle this outside the libhttpserver, but then what's the point in having `register_resource` at all?

### Steps to Reproduce

Build a server with the above code.

**Expected behavior:** [What you expect to happen]

Family and non-family resources with the same path should not conflict.

**Actual behavior:** [What actually happens]

Regardless of family/non-family, resources with the same path conflict.

**Reproduces how often:** [What percentage of the time does it reproduce?]

100%

### Versions

* All OS
* current  libhttpserver master branch
* libmicrohttpd version (most likely irrelevant)

PR #306, opened 2023-04-02
Consider family_url in ordering of http_endpoint.

bcsgh:
This will allow calling register_resource independently for a path as a family and non-family.

### Identify the Bug

https://github.com/etr/libhttpserver/issues/305

### Description of the Change

Consider family_url in ordering of http_endpoint.

### Alternate Designs

None.

### Possible Drawbacks

This will change behavior of programs in some cases. In general however, those cases are already not doing what the user likely expects.

### Verification Process

TBD

### Release Notes

Allow register_resource to use the same path for family and non-family cases in the same webserver.

### DIFF
--- a/src/details/http_endpoint.cpp
+++ b/src/details/http_endpoint.cpp
@@ -142,6 +142,7 @@ http_endpoint& http_endpoint::operator =(const http_endpoint& h) {
 }
 
 bool http_endpoint::operator <(const http_endpoint& b) const {
+    if (family_url != b.family_url) return family_url;
     COMPARATOR(url_normalized, b.url_normalized, std::toupper);
 }
 
--- a/test/integ/basic.cpp
+++ b/test/integ/basic.cpp
@@ -394,7 +394,7 @@ LT_BEGIN_AUTO_TEST(basic_suite, duplicate_endpoints)
     LT_CHECK_EQ(false, ws->register_resource("OK/", &ok2));
 
     // Check how family interacts.
-    LT_CHECK_EQ(false, ws->register_resource("OK", &ok2, true));
+    LT_CHECK_EQ(true, ws->register_resource("OK", &ok2, true));
 
     // Check that switched case does the right thing, whatever that is here.
 #ifdef CASE_INSENSITIVE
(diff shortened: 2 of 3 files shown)


src/details/http_endpoint.cpp
```diff
@@ -142,6 +142,7 @@ http_endpoint& http_endpoint::operator =(const http_endpoint& h) {
 }
 
 bool http_endpoint::operator <(const http_endpoint& b) const {
+    if (family_url != b.family_url) return family_url;
```
etr, line 145: Makes sense. I think we'd benefit from a few tests showing the behavior of family covering for already registered resourced while also showing that the specific resources are still called properly.

bcsgh: Figuring out how to do that is why this PR is still a draft. I think I've also found another spot where I need to tweak things a bit to make this correct. (In`webserver::register_resource` don't insert into `registered_resources_str` if `family`, but again, I still don't have tests for that.)

One things I'm running into re: tests is that all the integration tests try to run on the same hard coded port and I keep getting `Unable to connect daemon to port: 8080`. I'm real tempted to try to make that auto select a port but I'm not sure how much I want to dig around to figure out how to get said port back out.

test/integ/basic.cpp
```diff
@@ -406,6 +406,100 @@ LT_BEGIN_AUTO_TEST(basic_suite, duplicate_endpoints)
...
+    string s;
+    CURL *curl = curl_easy_init();
+    CURLcode res;
+    curl_easy_setopt(curl, CURLOPT_URL, "localhost:" PORT_STRING "/OK");
+    curl_easy_setopt(curl, CURLOPT_HTTPGET, 1L);
+    curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, writefunc);
+    curl_easy_setopt(curl, CURLOPT_WRITEDATA, &s);
+    res = curl_easy_perform(curl);
+    LT_ASSERT_EQ(res, 0);
+    LT_CHECK_EQ(s, "1");   // Not sure why regex wins, but it does...
```
etr, line 426: Regex just happens to replace the other. We do not advertise a specific behavior in the API interface on what wins in case of collisions (maybe we should).

I would just state the comment factually: "In the current implementation the regex wins".

bcsgh: Oops. That comment is actually wrong/irrelevant, it was copy/pasted from another test.

etr:
Thanks for the change. This is really helpful

PR #307, opened 2023-04-04
Modify some integration tests to allow them to pass on more limited systems

bcsgh:
Several integration tests require that port 8080 be available. As this is one of the more common alternative HTTP ports, it is likely to be unavailable. A bit of pre-processor hacking allows the build to select a different port if needed.

Another issue is that tests that use `httpserver::file_response` require that the expected path be an actual file. Some build environments (the provoking case is where I need to use [libhttpserver from a Bazel build](https://github.com/bcsgh/bazel_rules/blob/bcsgh/libhttpserver-tests/repositories/BUILD.libhttpserver), which uses a sandbox built from symlinks, and I need to also build and run the test under bazel to ensure that build is *also* correct)

---

*NOTE:* I don't know how to make Codacy/cppcheck happy with the new config `HTTPSERVER_PORT` (it seems to already be fine with the other config: `HTTPSERVER_NO_LOCAL_FS`) . The most useful thing I've found is [this](https://stackoverflow.com/questions/64567248/configurationnotchecked-error-with-cppcheck) but I don't know where to go from there.

Help or suggestions would be appreciated.

### DIFF
--- a/test/integ/basic.cpp
+++ b/test/integ/basic.cpp
@@ -316,7 +316,7 @@ class print_response_resource : public http_resource {
 
 #define STR2(p) #p
 #define STR(p) STR2(p)
-#define PORTS STR(PORT)
+#define PORT_STRING STR(PORT)
 
 LT_BEGIN_SUITE(basic_suite)
     webserver* ws;
@@ -347,7 +347,7 @@ LT_BEGIN_AUTO_TEST(basic_suite, two_endpoints)
     {
         CURL *curl = curl_easy_init();
         CURLcode res;
-        curl_easy_setopt(curl, CURLOPT_URL, "localhost:" PORTS "/OK");
+        curl_easy_setopt(curl, CURLOPT_URL, "localhost:" PORT_STRING "/OK");
         curl_easy_setopt(curl, CURLOPT_HTTPGET, 1L);
         curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, writefunc);
         curl_easy_setopt(curl, CURLOPT_WRITEDATA, &s);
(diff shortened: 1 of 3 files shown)


test/integ/basic.cpp
```diff
@@ -304,11 +308,21 @@ class print_response_resource : public http_resource {
...
 
+#ifdef HTTPSERVER_PORT
+#define PORT HTTPSERVER_PORT
+#else
+#define PORT 8080
+#endif  // PORT
+
+#define STR2(p) #p
+#define STR(p) STR2(p)
+#define PORTS STR(PORT)
```
etr, line 319: I'd just call this PORT_STRING as not to create confusion.

test/integ/basic.cpp
```diff
@@ -304,11 +308,21 @@ class print_response_resource : public http_resource {
...
 };
 
+#ifdef HTTPSERVER_PORT
+#define PORT HTTPSERVER_PORT
+#else
+#define PORT 8080
+#endif  // PORT
+
+#define STR2(p) #p
+#define STR(p) STR2(p)
```
etr, line 318: Why do we need the proxy definition of STR(p)? Cannot we just simply have what's on line 317 and name that STR(p) ?

bcsgh: Without the extra layer, the macro expands to the literal argument (`PORT`) rather than what that argument expands to:

A quick demo of how things expand: https://godbolt.org/z/Mz4eexzb5

See also: https://gcc.gnu.org/onlinedocs/gcc-4.8.5/cpp/Stringification.html

bcsgh:
Looks like codacy caches config per commit rather than per PR, adding a new commit seems to make it start over. :-)

test/integ/basic.cpp
etr: Makes sense, thanks.

etr:
Looks good. Merged it in.

Issue #308, opened 2023-04-06
[BUG] URL's that exactly match a registried regex pattern dispatch to that registered resource


bcsgh:
### Prerequisites

* [X] Put an X between the brackets on this line if you have checked that your issue isn't already filed: https://github.com/search?l=&q=repo%3Aetr%2Flibhttpserver&type=Issues

### Description

[This test here](https://github.com/etr/libhttpserver/blob/914ac9e6d2483347d11fc8bab0d85c9df1dc8b0e/src/webserver.cpp#L637-L638) is in error because that map is [populated with the URL including the regexp pattern](https://github.com/etr/libhttpserver/blob/914ac9e6d2483347d11fc8bab0d85c9df1dc8b0e/src/webserver.cpp#L201-L203).

This results in a url that includes the same text as the regex pattern matching, even if that text isn't matched by the regex.

### Steps to Reproduce

Add the following to `test/integ/basic.cpp` (https://github.com/etr/libhttpserver/pull/309/commits/2ea297f680f63649715eababf2badb824e467638):

```
LT_BEGIN_AUTO_TEST(basic_suite, regex_url_exact_match)
    ok_resource resource;
    ws->register_resource("/foo/{v|[a-z]}/bar", &resource);
    curl_global_init(CURL_GLOBAL_ALL);

    string s;
    CURL *curl = curl_easy_init();
    CURLcode res;

    // Exact string match for the registered pattern 
    curl_easy_setopt(curl, CURLOPT_URL, "localhost:8080/foo/{v|[a-z]}/bar/");

    curl_easy_setopt(curl, CURLOPT_HTTPGET, 1L);
    curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, writefunc);
    curl_easy_setopt(curl, CURLOPT_WRITEDATA, &s);
    res = curl_easy_perform(curl);
    LT_ASSERT_EQ(res, 0);

    int64_t http_code = 0;
    curl_easy_getinfo(curl, CURLINFO_RESPONSE_CODE , &http_code);
    LT_ASSERT_EQ(http_code, 404);    /// <<<<<<<<<<<<< This fails: 200!=404
    curl_easy_cleanup(curl);
LT_END_AUTO_TEST(regex_url_exact_match)
```

**Expected behavior:** HTTP 404 (assuming the RE pattern doesn't match itself).

**Actual behavior:** HTTP 200

**Reproduces how often:** 100%

### Versions

[...]

PR #309, opened 2023-04-06
Expand tests of resource resolution

bcsgh:
Add some more test for resource resolution.

This includes
- a test exposing a bug: https://github.com/etr/libhttpserver/issues/308
- some tests of some of the corners relating to another PR: https://github.com/etr/libhttpserver/pull/306

### DIFF
--- a/test/integ/basic.cpp
+++ b/test/integ/basic.cpp
@@ -386,7 +386,7 @@ LT_BEGIN_AUTO_TEST(basic_suite, duplicate_endpoints)
 #ifdef CASE_INSENSITIVE
     LT_CHECK_EQ(false, ws->register_resource("ok", &ok2));
 #else
-    // TODO: this should be true.
+    // TODO(etr): this should be true.
     // However, http_endpoint::operator< is always case-insensitive
     LT_CHECK_EQ(false, ws->register_resource("ok", &ok2));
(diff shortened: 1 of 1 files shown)


test/integ/basic.cpp
```diff
@@ -358,6 +369,68 @@ LT_BEGIN_AUTO_TEST(basic_suite, two_endpoints)
...
+
+    // Check how family interacts.
+    LT_CHECK_EQ(false, ws->register_resource("OK", &ok2, true));
+
+    // Check that switched case does the right thing, whatever that is here.
+#ifdef CASE_INSENSITIVE
+    LT_CHECK_EQ(true, ws->register_resource("ok", &ok2));
+#else
+    LT_CHECK_EQ(false, ws->register_resource("ok", &ok2));
+#endif
```
bcsgh, line 390: After a bit more digging, I think these are actually backwards. And both in fact end up returning false.

tl;dr: `http_endpoint::operator <` is unconditionally case-insensitive:
https://github.com/etr/libhttpserver/blob/6d1df063501c9a7b75d68958fa3105a154a75cf0/src/details/http_endpoint.cpp#L144-L146 

That's going to be a pain to fix as it's something users could be inadvertently depending on.

PR #312, opened 2023-04-13
Make more robust tests.

bcsgh:
This is "more of the same" with regards to making the tests more robust about the environment they are run in.

- Apply the `HTTPSERVER_PORT` patch everywhere.
- Do the same pattern for `HTTPSERVER_DATA_ROOT` for loading test files.
- Add some tests for things that silently failed and showed up while debugging some of the above tweaks.
- Add some tests for things that didn't cause the above but could have been issues.
- Some misc cleanup.

### DIFF
--- a/test/integ/authentication.cpp
+++ b/test/integ/authentication.cpp
@@ -99,7 +99,7 @@ LT_BEGIN_AUTO_TEST(authentication_suite, base_auth)
     webserver ws = create_webserver(PORT);
 
     user_pass_resource user_pass;
-    ws.register_resource("base", &user_pass);
+    LT_ASSERT_EQ(true, ws.register_resource("base", &user_pass));
     ws.start(false);
 
     curl_global_init(CURL_GLOBAL_ALL);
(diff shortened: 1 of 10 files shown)


bcsgh:
@etr Can you take a look at the CI failures? I'm unable to reproduce that specific error locally and I *think* the one in CI might be pre-existing? (It seems that exceptions escaping a test are silently ignored?)

etr:
It seems to be impacting only the SSL related tests and I agree with you, this seems to be pre-existing.

I think as we moved to github-actions, we missed the installation of the gnutls dependency that libmicrohttpd needs to enable SSL.

I need to check why the tests weren't failing though, I suspect that this is because the exception is not on the main thread, but nonetheless, it would be worth for this to be caught.

etr:
The issue was due to libgnutls not being installed in the CI flow. Now I still have one remaining issue with one of the tests legitimately failing.

etr:
It should all work now if you pull from live

bcsgh:
> It should all work now if you pull from live

Am I still doing something wrong?

```
[ASSERT FAILURE] (../../test/integ/ws_start_stop.cpp:460) - error in "ssl_with_protocol_priorities": exceptions thown by ((ws.start(false)))
```

Looking at the diff it *look* like I've correctly propagated your fix:

```C++
LT_ASSERT_NOTHROW(cws.https_priorities("NONE:+AES-256-GCM:+SHA384"));
```

bcsgh:
Wrapping individual lines in `LT_ASSERT_NOTHROW` seems like a sisyphean task, I've backed out some of that in favor of solving that at the more general layer: https://github.com/etr/libhttpserver/pull/316

PR #313, opened 2023-04-13
Test to see if CI fails for everything.

bcsgh:
DO NOT REVIEW

Just checking that https://github.com/etr/libhttpserver/issues/314 is real and not something I broke.


etr:
Codecov :( - the integration broke completely before I could get around to fix the integration (they've removed the package). Will remove it for now and re-build the coverage properly.

PR #316, opened 2023-04-15
Count (unexpected) escaping exceptions as a test failure.

bcsgh:

https://github.com/etr/liblittletest/issues/3

### Description of the Change

Make LittleTest count exceptions escaping from any of the set-up, tear-down or test-run (other than asserts) as a test failure. 

### Alternate Designs

None, considered.

### Possible Drawbacks

This could cause existing test to start failing, But (fingers crossed) that's only likely for test that should already be failing.

### Verification Process

- Ran the `libhttperver` tests and found another cases I didn't already know about that was failing for me.
- The CI for this PR fails as a result of a pre-existing silent error.

### Release Notes

Make LittleTest count exceptions escaping from any of the set-up, tear-down or test-run (other than asserts) as a test failure. 

### See also:

- https://github.com/etr/liblittletest/pull/4

### DIFF
--- a/.github/workflows/verify-build.yml
+++ b/.github/workflows/verify-build.yml
@@ -371,14 +371,13 @@ jobs:
     - name: Setup dependencies (only on linux)
       run: |
         sudo apt-get install info install-info ;
-        sudo pip install gcovr ;
         sudo apt-get install cppcheck ;
       if: ${{ matrix.os-type == 'ubuntu' }}
 
     - name: Setup coverage dependencies (only on linux)
       run: |
         sudo pip install gcovr ;
-      if: ${{ matrix.os-type == 'ubuntu' && matrix.c-compiler == 'gcc' && matrix.debug == 'debug' && matrix.coverage == 'coverage' && success() }}
+      if: ${{ matrix.os-type == 'ubuntu' && matrix.c-compiler == 'gcc' && matrix.debug == 'debug' && matrix.coverage == 'coverage' }}
 
     - name: Setup gnutls dependency (only on linux)
       run: |
--- a/.github/workflows/verify-build.yml
+++ b/.github/workflows/verify-build.yml
@@ -368,12 +368,17 @@ jobs:
       run: brew install autoconf automake libtool
       if: ${{ matrix.os == 'macos-latest' }}
 
-    - name: Setup coverage dependencies (only on linux)
+    - name: Setup dependencies (only on linux)
       run: |
         sudo apt-get install info install-info ;
         sudo pip install gcovr ;
         sudo apt-get install cppcheck ;
       if: ${{ matrix.os-type == 'ubuntu' }}
+
+    - name: Setup coverage dependencies (only on linux)
+      run: |
+        sudo pip install gcovr ;
+      if: ${{ matrix.os-type == 'ubuntu' && matrix.c-compiler == 'gcc' && matrix.debug == 'debug' && matrix.coverage == 'coverage' && success() }}
 
     - name: Setup gnutls dependency (only on linux)
       run: |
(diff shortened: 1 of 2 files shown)


bcsgh:
Bump?

I think the CI errors were already there, the test was already broken, it just wasn't being reported as such.

etr:
I did not have enough time to investigate this, unfortunately.

It appears to be on the setting of protocol priorities for HTTPS. But I need to find the time to run some local tests to see why it keeps failing,

bcsgh:
IIRC I came to the same onclusion about where the issue was. 

I think I ended up using this patch to debug things. https://github.com/bcsgh/libhttpserver/commit/acbd272b78ec441e62891c27895644c309bf2104

bcsgh:
Rebasing on your https://github.com/etr/libhttpserver/commit/82ae1f088e9e3369a0b2de8d12cfc126200fd05a seems to resolve those prior issues, but now coverage seems to be broken?

```
Collecting gcovr
  ERROR: HTTP error 502 while getting https://files.pythonhosted.org/packages/11/e9/4eb38ed44d7485a6343f24db3cfc773362a50f3413d7c11678cac2ec6dec/gcovr-6.0-py2.py3-none-any.whl#sha256=2e52019fdb76c6e327f48c2a2d8555fb5e362570b79cc74c5498804d1ce54a60
(from https://pypi.org/simple/gcovr/) (requires-python:>=3.7)
```

Given that `gcovr` has needed  >=3.7 for over a year I'm guessing something else has changed?

The offending URL has `py2` in it which kinda suggests that switching from `pip install` to `pip3 install` [here](https://github.com/etr/libhttpserver/blob/master/.github/workflows/verify-build.yml#L374) might fix things? (Don't know, I don't have time to look at it this morning.)

bcsgh:
Not sure if it was just re-running the tests (I've not managed to repro that fail) or adding https://github.com/bcsgh/libhttpserver/commit/bb132826accf2f3963f03fd6d10b1d41d1e56dd7, but tests are now clean.

etr:
> I've not managed to repro that fail

You mean the failure on the HTTPS priorities? I suspect that setting a set of priorities that made sense actually helped.

bcsgh:
What I've failed to repro is this failure of `pip install gcovr`: https://github.com/etr/libhttpserver/actions/runs/4852163596/jobs/8646809985

.github/workflows/verify-build.yml
```diff
@@ -368,13 +368,18 @@ jobs:
...
 
-    - name: Setup coverage dependencies (only on linux)
+    - name: Setup dependencies (only on linux)
       run: |
         sudo apt-get install info install-info ;
         sudo pip install gcovr ;
         sudo apt-get install cppcheck ;
       if: ${{ matrix.os-type == 'ubuntu' }}
 
+    - name: Setup coverage dependencies (only on linux)
```
etr, line 378: I don't suspect this is going to do much as it is since gcovr is already installed in the previous step.

It might be worth removing gcovr from the previous step if we are anyway restricting it here (which is fine).

The other thing is the "success()" in the if statement. I think that will wait for the overall run to succeed before installing gcovr - I don't think that should be there.

bcsgh: Done.

That said, I think this ran correctly even with the `&&success()`. (Does that check that all *prior* steps passed? If not, how would it know what steps are "the entire run"?)

etr:
You are right, and my github-actions-fu is weak.

See: https://docs.github.com/en/actions/learn-github-actions/expressions#success

etr:
I will likely back port some of your changes into: https://github.com/etr/liblittletest

bcsgh:
> I will likely back port some of your changes into: https://github.com/etr/liblittletest

https://github.com/etr/liblittletest/pull/4/files

(There are a few other deltas between the two files that could be cross merged, but that PR should cover the stuff added in the PR.)

PR #319, opened 2023-05-02
DO NOT MERGE

bcsgh:
Just checking if the test pass.

If this works, then once https://github.com/etr/libhttpserver/pull/316 is merged I'll point https://github.com/etr/libhttpserver/pull/312 here and send it for review.


bcsgh:
Done with this.

PR #320, opened 2023-05-02
Added a get-method for the content of a string_response

hsiegmund:
### Description of the Change

Added a getter for the content of a string_response.
Example usage: a custom logger that shall log the content.

### Release Notes

Added a get-method for the content of a string_response.


etr:
I am not against having a get_content (or something of the likes) as part of the `http_response`, possibly to benefit logging. In the current approach though, this covers only `string_response` and I am not sure why this variant of the `http_response` should have different treatment from the rest.

I would propose we build something that can be reuse across all types of responses.

PR #322, opened 2023-05-14
A bunch of generalized code cleanup.

bcsgh:
Mostly proactive bug-avoidance.

### DIFF
--- a/src/http_request.cpp
+++ b/src/http_request.cpp
@@ -35,8 +35,8 @@ struct arguments_accumulator {
     std::map<std::string, std::vector<std::string>, http::arg_comparator>* arguments;
 };
 
-void http_request::set_method(const std::string& method_) {
-    method = string_utilities::to_upper_copy(method_);
+void http_request::set_method(const std::string& method) {
+    this->method = string_utilities::to_upper_copy(method);
 }
 
 bool http_request::check_digest_auth(const std::string& realm, const std::string& password, int nonce_timeout, bool* reload_nonce) const {
--- a/src/http_request.cpp
+++ b/src/http_request.cpp
@@ -35,8 +35,8 @@ struct arguments_accumulator {
     std::map<std::string, std::vector<std::string>, http::arg_comparator>* arguments;
 };
 
-void http_request::set_method(const std::string& method) {
-    this->method = string_utilities::to_upper_copy(method);
+void http_request::set_method(const std::string& method_) {
+    method = string_utilities::to_upper_copy(method_);
 }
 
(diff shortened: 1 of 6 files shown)


src/http_request.cpp
```diff
@@ -35,8 +35,8 @@ struct arguments_accumulator {
     std::map<std::string, std::vector<std::string>, http::arg_comparator>* arguments;
 };
 
-void http_request::set_method(const std::string& method) {
-    this->method = string_utilities::to_upper_copy(method);
+void http_request::set_method(const std::string& method_) {
+    method = string_utilities::to_upper_copy(method_);
```
etr, line 39: I think I prefer the "this->" in setters/grow methods. Probably a matter of taste, but I find it more explicit.

bcsgh: In my experience `this->` generally *screams* "there's something strange going on here!"

The only thing at all strange here is having a method argument shadow an object member (which I think is a minor issue all on its own). Both the name of the method and the thing being assigned to call out what it is, so I'm tempted to propose just abbreviating the argument down to a single letter?

Other than shadowing, I don't see any argument for being explicit here that doesn't apply every place a method or member is accessed. Which is to say, if you don't want that other places; why here?

etr: As said, it is a matter of taste.

> In my experience this-> generally screams "there's something strange going on here!"

It doesn't really scream anything to me.

> Both the name of the method and the thing being assigned to call out what it is, so I'm tempted to propose just abbreviating the argument down to a single letter?

No, I prefer keeping the semantic meaning of variables. Agree with you that C++ has the unfortunate property of being verbose in its setters, but after all, we write it once and I don't find it hurts my ability to read the setter. If anything, if I see a different variable name there, that's where I'd pause.

> Other than shadowing, I don't see any argument for being explicit here that doesn't apply every place a method or member is accessed. Which is to say, if you don't want that other places; why here?

I generally don't care if it is used elsewhere, but I find it slightly redundant when used for other purposes outside of shadowing.

In this case, we are indeed shadowing, so I find it quite appropriate to use. And given that the alternative is to use a conventional "_" after the variable name, I don't think we are achieving much. We are just using a convention over a proper tool that the language has given us to solve the shadowing.

bcsgh: You're the owner so I'll respect your preferences.

That said, the issue of shadowing (as I'm intending the term) isn't fixed by qualifying the access. The issue is that there are multiple things that an unqualified reference could resolve to. This issue manifests, for example, if somehow the argument name got changed and suddenly the references resolve to something different than they did before. That results in a sort of "spooky action at a distance" that can be hard to debug, or even notice.

Issue #329, opened 2023-07-24
[BUG] Data races


FlorianChevassu:
### Prerequisites

* [X] Put an X between the brackets on this line if you have checked that your issue isn't already filed: https://github.com/search?l=&q=repo%3Aetr%2Flibhttpserver&type=Issues

### Description

The documentation states that `All functions are guaranteed to be completely reentrant and thread-safe (unless differently specified)`. However, the `webserver` implementation does not use any protection while reading/modifying its member variables (`registered_resources`, `registered_resources_str`, `bans` and `allowances`).

### Steps to Reproduce

Here is a small test that show the issue when compiled using clang-16 with thread sanitizing enabled:
```cpp

LT_BEGIN_AUTO_TEST(basic_suite, thread_safety)
    simple_resource resource;

    std::atomic_bool done = false;
    auto register_thread = std::thread([&]() {
        int i = 0;
        using namespace std::chrono;
        while (!done) {
            ws->register_resource(
                    std::string("/route") + std::to_string(++i), &resource);
        }
    });

    auto get_thread = std::thread([&](){
        while (!done) {
            CURL *curl = curl_easy_init();
            std::string s;
            std::string url = "localhost:" PORT_STRING "/route" + std::to_string(
                                            (int)((rand() * 10000000.0) / RAND_MAX));
            curl_easy_setopt(curl, CURLOPT_URL, url.c_str());
            curl_easy_setopt(curl, CURLOPT_HTTPGET, 1L);
            curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, writefunc);
            curl_easy_setopt(curl, CURLOPT_WRITEDATA, &s);
            curl_easy_perform(curl);
            curl_easy_cleanup(curl);
        }
    });

    using namespace std::chrono_literals;
    std::this_thread::sleep_for(10s);
    done = true;
    if (register_thread.joinable()) {
        register_thread.join();
    }
    if (get_thread.joinable()) {
        get_thread.join();
    }
    LT_CHECK_EQ(1, 1);
LT_END_AUTO_TEST(thread_safety)
```

**Expected behavior:** 

No data races are detected by clang.

**Actual behavior:**
Clang reports the following data race (and others):

```txt
Running test (1): thread_safety
==================
WARNING: ThreadSanitizer: data race (pid=21795)
  Read of size 8 at 0x7b1400003550 by thread T1:
[...]

PR #330, opened 2023-07-25
Fix webserver thread safety

FlorianChevassu:
### Identify the Bug

https://github.com/etr/libhttpserver/issues/329

### Description of the Change

This pull request introduces the usage of locks in the werbserver class in order to protect member variables from data races.

### Alternate Designs

N/A

### Possible Drawbacks

This might slightly increase response time, but should not be noticeable.

### Verification Process

N/A

### Release Notes

Fixed an issue where multiple threads did access webserver member variables concurrently, leading to data races.


src/httpserver/webserver.hpp
```diff
@@ -167,16 +168,18 @@ class webserver {
...
-     render_ptr method_not_allowed_resource;
-     render_ptr internal_error_resource;
+     const render_ptr not_found_resource;
+     const render_ptr method_not_allowed_resource;
+     const render_ptr internal_error_resource;
+     std::shared_mutex registered_resources_mutex;
      std::map<details::http_endpoint, http_resource*> registered_resources;
      std::map<std::string, http_resource*> registered_resources_str;
 
+     std::shared_mutex bans_and_allowances_mutex;
```
etr, line 182: It might be worth splitting this into two, given the writing use-cases are isolated

src/webserver.cpp
```diff
@@ -402,6 +409,7 @@ MHD_Result policy_callback(void *cls, const struct sockaddr* addr, socklen_t add
 
     if (!(static_cast<webserver*>(cls))->ban_system_enabled) return MHD_YES;
 
+    std::shared_lock bans_and_allowances_lock((static_cast<webserver*>(cls))->bans_and_allowances_mutex);
```
etr, line 412: We can split the if statement here so that we always check only one of the two based on the chosen policy

etr:
Overall, it looks good - just a few minor comments

src/httpserver/webserver.hpp
FlorianChevassu: done

src/webserver.cpp
FlorianChevassu: In both ACCEPT and REJECT cases, we access `bans` and `allowances`, so I don't see how we could only lock one of the mutexes.

I added a commit simplifying the code, with no intended functional change.

etr: [MINOR] I think the cleanest way would be to have two private methods (is_allowed and is_banned) that use the locks internally and just call into those methods from within here. That should localize the locks to the maximum extent possible.

etr:
It looks like the PR is failing at compiling

FlorianChevassu:
> It looks like the PR is failing at compiling

Sorry about the delay. Should be fixed now.

PR #333, opened 2024-03-02
Adding TLS-PSK authentication method via function callback.

oberdorc:
### Identify the Bug adding a Feature

- There was a bug int the automake configuration file. The ' HAVE_GNUTLS' tag is never defined in the compilation although presence of TLS header files has been verified.
-  Corrected for an error message by _libmicrohttpd_ that complains about possibly missing log messages if an external logger is not assigned immediately set as the first option.
- Eventually, added callback function for TLS-PSK authentication.

### Description of the Change

- The configuration file has been updated to correct for the aforementioned bug. I also run autoupdate on the configuration file to deal with deprecated behavior on STDC_HEADER.
- The order of options in the list of configuarion options has been changed (iov...).
- Remaining changes circle around adding PSK authentication:
1. Added option to set a callback function of type _psk_cred_handler_callback_ for PSK authentication to create_webserver.hpp. The function is wrapped inside a std::function and stored as a value type in the create_webserver instance (no need for pointers with std::function IMO. Maybe this should be changed for the other  callback function handlers there as well, e.g. log_access_ptr, ... ).
2. Added a static wrapper function _psk_cred_handler_func_ that will be called by the _libmicrohttpd_ library as the PSK callback and that will in turn call the dedicated callback function from _libhttpserver_. This function converts the original hex-encoded PSK to the binary key sequence required by _libmicrohttpd_ by calling gnutls_hex2bin from the gnutls library.
3. PSK authentication works with and without providing a certificate and key. Checks during the initialization of the webserver have been changed accordingly. SSL related options are initialized if SSL is enabled and a non empty option value has been assigned.
4. Eventually, the README.md has been edited to account for the added PSK authentication feature.

### Verification Process

Obviously without HAVE_GNUTLS beeing defined essential parts of the code are not compiled. This has been checked making sure that the flag is included in the config.h file that is created for the build process.

The error message from _libmicrohttpd_ has dissappeared.

A mininimal example and test case for HTTPS PSK authentication has been added and test with the openssl command.


oberdorc:
So, there remains one formatting issue indicated by cpplint, which seems to be a false-positive. Quite sad. And then there is some issue with a conftest program which fails to compile for some reason. I give up on that.

